PR #61, opened 2018-10-16
Fix 52: "EEXIST:file already exists" raises during creating java project

### LINKED ISSUE #52: Create java project error "EEXIST:file already exists, mkdir 'd:\vscode-java-dependency\demo2'"
Build: 41
Plat: windows/Mac/Linux

Steps:

1. F1, open the command palette
2. Select Java: Create Java Project
3. Select the location
4. Input the project name

The result  as below:

image
The created folder details as below:

image

The normal one as below:

image

andxu:
…project, using Promise.all will cause the racing of creating the target folder.

Fix #52


src/controllers/projectController.ts
```diff
@@ -53,12 +53,11 @@ export class ProjectController {
         const projectFile: string = path.join(basePath, projectName, ".project");
         const extensionPath: string = this.context.extensionPath;
         try {
-            await Promise.all([
```
SummerSun: Maybe ensure the directory first and put all the copy work in one Promise.all()?

src/controllers/projectController.ts
```diff
@@ -53,12 +53,11 @@ export class ProjectController {
...
-                fse.copy(path.join(extensionPath, "templates", "App.java.sample"), path.join(basePath, projectName, "src", "app", "App.java")),
-                fse.copy(path.join(extensionPath, "templates", `Java${javaVersion}`), path.join(basePath, projectName)),
-                fse.copy(path.join(extensionPath, "templates", ".project"), projectFile),
-                fse.ensureDir(path.join(basePath, projectName, "bin")),
-            ]);
+            await fse.ensureDir(path.join(basePath, projectName, "bin"));
```
akaroml: These two has very different meanings. The later makes it a serial operation. I don't see a problem with Promise.all. Please keep that one.

src/controllers/projectController.ts
akaroml: I agree with @SummerSun and disagree with the proposed solution. Promise.all is well known so it does not make things harder to understand. The performance downgrade is definitely noticeable when you work with larger projects. Triggering multiple copy task and let them run in parallel will make the most of modern hardware like SSD. Running them in serial manner does not make any sense.

templates/App.java.sample
```diff
@@ -1,8 +1,7 @@
 package app;
 
 public class App {
-	public static void main(String[] args) throws Exception {
-		System.out.println("Hello Java");
-	}
-
+    public static void main(String[] args) throws Exception {
+        System.out.println("Hello Java");
```
SummerSun: Any change of this code snippet?

src/controllers/projectController.ts
```diff
@@ -50,14 +50,17 @@ export class ProjectController {
     }
 
     private async scaffoldJavaProject(basePath: string, projectName: string, javaVersion: number): Promise<boolean> {
-        const projectFile: string = path.join(basePath, projectName, ".project");
-        const extensionPath: string = this.context.extensionPath;
+        const projectRoot: string = path.join(basePath, projectName);
+        const templateRoot: string = path.join(this.context.extensionPath, "templates");
+        const projectFile: string = path.join(projectRoot, ".project");
         try {
+            fse.ensureDirSync(projectRoot);
```
SummerSun: Nice!

templates/App.java.sample
andxu: Replacing tab with spaces

src/controllers/projectController.ts
akaroml: Can we just await fse.ensureDir?
A sync operation blocks others especially when there is potential retry logic.

PR #68, opened 2018-10-17
fix outline after lsp4j api change

andxu:


andxu:
https://github.com/Microsoft/vscode-java-dependency/issues/69

Issue #57, opened 2018-10-16
Should have a option for hierarchical tree view other than flat view for package.


andxu:

Flanker32:
Fixed in pr #117

PR #117, opened 2018-11-30
Add hierarchical package view, for issue #57

Flanker32:
Add hierarchical package view, for issue #57


package.json
```diff
@@ -62,6 +62,11 @@
           "type": "boolean",
           "description": "Synchronize dependency viewer selection with folder explorer",
           "default": true
+        },
+        "java.dependency.isHierarchicalView": {
```
yaohaizh: Should avoid use this value.  It is not clear to the end user what the opposite is? 

Should be enum values: "flat", "hierarchy"

src/views/dependencyExplorer.ts
```diff
@@ -49,7 +50,6 @@ export class DependencyExplorer {
         if (!current) {
             return;
         }
-
```
yaohaizh: Avoid the empty diff if possible. You can always provide another PR for format/coding convention to avoid distraction.

src/views/packageRootNode.ts
```diff
@@ -14,15 +16,44 @@ import { TypeRootNode } from "./typeRootNode";
 
 export class PackageRootNode extends DataNode {
 
-    constructor(nodeData: INodeData, parent: DataNode, private _project: ProjectNode) {
+    private packageTree: PackageTreeNode;
+
+    constructor(nodeData: INodeData, parent: DataNode, private _project: ProjectNode, packageTree: PackageTreeNode = null) {
         super(nodeData, parent);
```
yaohaizh: You should NOT mixed the flat package node and hierarchy package node in the same class. It violate SRP and will make the code hard to main in the future. 

You can create two class/types for different purpose.

package.json
Flanker32: Resolved

src/views/packageRootNode.ts
Flanker32: I made an abstract class named `HierarchicalNode` , which includes `createFlatChildNodeList` , `createHierarchicalChildNodeList` and `revealPath`, and `PackageRootNode` extends it, i didn't split packagenode into two parts for there are still some common methods like loadData and getIconPath, other methods for Hierarchical was moved to `HierachicalPackageRootSubNode`

src/views/dependencyExplorer.ts
```diff
@@ -71,8 +71,14 @@ export class DependencyExplorer {
                             this._selectionWhenHidden = c;
                         }
                     } else {
-                        paths.shift();
-                        this.revealPath(c, paths);
+                        // Resove Hierarchical packages
```
yaohaizh: typo: Resolve ==> Resolve

src/settings.ts
```diff
@@ -28,5 +28,9 @@ export class Settings {
         return this._depdendencyConfig.get("syncWithFolderExplorer");
     }
 
+    public static isHierarchicalView(): boolean {
+        return this._depdendencyConfig.get("packagePresentation") === "hierarchical";
```
andxu: should avoid to use the configuration directly since the config may change at runtime, in your impl, change of this options will  take effective on next update. see https://github.com/Microsoft/vscode-java-debug/blob/bcd69b52e4d941323cae662a154afd2a7852f43d/src/debugCodeLensProvider.ts#L38

package.json
```diff
@@ -62,6 +62,12 @@
           "type": "boolean",
           "description": "Synchronize dependency viewer selection with folder explorer",
           "default": true
+        },
+        "java.dependency.packagePresentation": {
+          "type": "string",
+          "enum":["flat","hierarchical"],
+          "description": "Set the presentation model of packages",
```
andxu: Package presentation mode: flat or hierarchical

src/java/packageTreeNode.ts
```diff
@@ -0,0 +1,66 @@
...
+        this.name = splitPackageName[0];
+        this.fullName = parentName === "" ? this.name : parentName + "." + this.name;
+        if (splitPackageName.length > 1) {
+            this.childs.push(new PackageTreeNode(packageName.substring(this.name.length + 1), this.fullName));
+        } else {
+            this.isPackage = true;
+        }
+    }
+
+    public addPackage(packageName: string): void {
```
andxu: addPackage constructor  compressTree should be better organized to clearance to detect tree structures based on a list of packages:
a.b
a.b.c.d
a.b.c.d.e

src/views/hierachicalPackageRootNode.ts
```diff
@@ -0,0 +1,97 @@
...
+import { HierachicalPackageRootSubNode } from "./hierachicalPackageRootSubNode";
+import { PackageRootNode } from "./packageRootNode";
+import { ProjectNode } from "./projectNode";
+import { TypeRootNode } from "./typeRootNode";
+
+export class HierachicalPackageRootNode extends PackageRootNode {
+
+    public static async convertPaths(paths: INodeData[]): Promise<INodeData[]> {
+        const index = paths.findIndex((nodeData) => nodeData.kind === NodeKind.PackageRoot);
+        const projectNodeData = paths.find((nodeData) => nodeData.kind === NodeKind.Project);
```
yaohaizh: All these operations are model oriented. You can compute the tree without creating a view. 

For example, you can make the packagetree as a the model and build the hierarchy model outof the the view.

src/java/packageTreeNode.ts
```diff
@@ -0,0 +1,65 @@
...
+    public isPackage: boolean = false;
+
+    private constructor(name: string, parentFullName: string) {
+        this.name = name;
+        this.fullName = parentFullName === "" ? name : parentFullName + "." + name;
+    }
+
+    public addPackage(packageName: string): void {
+        const packages = packageName.split(".");
+        this.addSubPackage(packages);
```
andxu: this.addSubPackage(packageName.split("."));

src/views/hierachicalPackageRootNode.ts
```diff
@@ -0,0 +1,71 @@
...
+                if (data.kind === NodeKind.File) {
+                    result.push(new FileNode(data, this));
+                } else if (data.kind === NodeKind.Folder) {
+                    result.push(new FolderNode(data, this, this._project, this));
+                } else if (data.kind === NodeKind.TypeRoot) {
+                    result.push(new TypeRootNode(data, this));
+                }
+            });
+        }
+        const packageNodeList = this.getHierarchicalPackageNodes();
```
andxu: don't introduce local var here

src/views/hierachicalPackageRootSubNode.ts
```diff
@@ -0,0 +1,63 @@
...
+import { FileNode } from "./fileNode";
+import { FolderNode } from "./folderNode";
+import { ProjectNode } from "./projectNode";
+import { TypeRootNode } from "./typeRootNode";
+
+export class HierachicalPackageRootSubNode extends DataNode {
+
+    public packageTree: PackageTreeNode;
+
+    constructor(nodeData: INodeData, parent: DataNode, private _project: ProjectNode, packageTree: PackageTreeNode = null) {
```
andxu: = null should be removed

src/views/dependencyDataProvider.ts
```diff
@@ -67,6 +68,37 @@ export class DependencyDataProvider implements TreeDataProvider<ExplorerNode> {
         return element.getParent();
     }
 
+    public async getRootNodeByData(nodeData: INodeData): Promise<DataNode> {
+        // Server only return project nodes, so use get root projects function
```
yaohaizh: As said, in order to get the data, creating the view is not an acceptable solution here.  DON't mix the model with your view. You either either directly get the data from the server, or patch the Workspsace data to the path side.

src/views/hierachicalPackageRootSubNode.ts
```diff
@@ -0,0 +1,63 @@
...
+import { DataNode } from "./dataNode";
+import { ExplorerNode } from "./explorerNode";
+import { FileNode } from "./fileNode";
+import { FolderNode } from "./folderNode";
+import { ProjectNode } from "./projectNode";
+import { TypeRootNode } from "./typeRootNode";
+
+export class HierachicalPackageRootSubNode extends DataNode {
+
+    public packageTree: PackageTreeNode;
```
yaohaizh: This class code is duplicate with the HierarchicalPackageRootNode, could we model these two types in one class?

src/java/packageTreeNode.ts
```diff
@@ -0,0 +1,65 @@
...
+            this.fullName = this.fullName + "." + child.name;
+            this.name = this.name + "." + child.name;
+            this.childs = child.childs;
+            this.isPackage = child.isPackage;
+        }
+        this.childs.forEach((child) => child.compressTree());
+    }
+
+    public getNodeDataFromPackageTreeNode(nodeData: INodeData): INodeData {
+        return {
```
yaohaizh: Typically, this is an adapter that we can apply here.

src/java/packageTreeNode.ts
```diff
@@ -0,0 +1,65 @@
...
+        this.name = name;
+        this.fullName = parentFullName === "" ? name : parentFullName + "." + name;
+    }
+
+    public addPackage(packageName: string): void {
+        const packages = packageName.split(".");
+        this.addSubPackage(packages);
+    }
+
+    public compressTree(): void {
```
yaohaizh: This is an internal operations. Should be private.

src/java/packageTreeNode.ts
yaohaizh: Second this. Currently, the compressTree will not work for this case.

src/views/hierachicalPackageRootSubNode.ts
```diff
@@ -0,0 +1,63 @@
...
+                }
+            });
+        }
+        const packageNodeList = this.getHierarchicalPackageNodes();
+        return packageNodeList.concat(result);
+    }
+
+    protected getHierarchicalPackageNodes(): ExplorerNode[] {
+        const result = [];
+        this.packageTree.childs.forEach((childNode) => {
```
yaohaizh: childs ==> children

src/views/hierachicalPackageRootNode.ts
```diff
@@ -0,0 +1,61 @@
...
+export class HierachicalPackageRootNode extends PackageRootNode {
+
+    constructor(nodeData: INodeData, parent: DataNode, _project: ProjectNode) {
+        super(nodeData, parent, _project);
+    }
+
+    public async revealPaths(paths: INodeData[]): Promise<DataNode> {
+        const hierachicalNodeData = paths[0];
+        const childs: ExplorerNode[] = await this.getChildren();
+        const childNode = <DataNode>childs.find((child: DataNode) =>
```
yaohaizh: Null reference check.

src/views/hierachicalPackageRootNode.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    protected createChildNodeList(): ExplorerNode[] {
+        const result = [];
+        if (this.nodeData.children && this.nodeData.children.length) {
+            this.sort();
+            this.nodeData.children.forEach((data) => {
+                if (data.kind === NodeKind.File) {
+                    result.push(new FileNode(data, this));
+                } else if (data.kind === NodeKind.Folder) {
```
yaohaizh: Call super methods first

src/views/hierachicalPackageRootNode.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    private getHierarchicalPackageNodeData(): HierachicalPackageNodeData {
+        if (this.nodeData.children && this.nodeData.children.length) {
+            const nodeDataList = this.nodeData.children
+                .filter((child) => child.kind === NodeKind.Package);
+            return HierachicalPackageNodeData.createHierachicalNodeDataByPackageList(nodeDataList);
+        } else {
+            // Return a empty hierachical node
+            return HierachicalPackageNodeData.createHierachicalNodeDataByPackageList([]);
```
yaohaizh: Duplicate call. You just need create package list with arguments. Not necessary two calls here.

Issue #139, opened 2018-12-27
Get NPE when opening a non-java file


jdneo:
### Repro steps
1. Open a non-java file, for example: Open the language server log file
2. You can find errors in the developer tools console panel.

### Traces
```
  ERR Cannot read property 'path' of undefined: TypeError: Cannot read property 'path' of undefined
	at module.exports.e.DependencyDataProvider.revealPaths.n.find.t (C:\Users\sheche\.vscode\extensions\vscjava.vscode-java-dependency-0.3.0\dist\extension.js:46:26196)
	at Array.find (native)
	at module.exports.e.DependencyDataProvider.<anonymous> (C:\Users\sheche\.vscode\extensions\vscjava.vscode-java-dependency-0.3.0\dist\extension.js:46:26177)
	at Generator.next (<anonymous>)
	at s (C:\Users\sheche\.vscode\extensions\vscjava.vscode-java-dependency-0.3.0\dist\extension.js:46:24568)
	at propagateAslWrapper (C:\Users\sheche\.vscode\extensions\vscjava.vscode-java-pack-0.5.0\node_modules\async-listener\index.js:504:23)
	at asyncWrap (C:\Users\sheche\.vscode\extensions\vscjava.vscode-java-pack-0.5.0\node_modules\async-listener\glue.js:188:31)
	at C:\Users\sheche\.vscode\extensions\vscjava.vscode-java-pack-0.5.0\node_modules\async-listener\index.js:541:70
	at asyncWrap (C:\Users\sheche\.vscode\extensions\vscjava.vscode-java-pack-0.5.0\node_modules\async-listener\glue.js:188:31)
	at <anonymous>
```

PR #141, opened 2018-12-27
FIx npe issue when open a non java file

Flanker32:
For issue #139


src/views/dependencyExplorer.ts
```diff
@@ -54,4 +57,5 @@ export class DependencyExplorer {
             this._selectionWhenHidden = node;
         }
     }
+
```
jdneo: remove this line

PR #172, opened 2019-08-15
Use maven wrapper to lock mvn version

Eskibear:
Azure pipeline fails due to Tycho bug in Maven 3.6.1. 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=546463

Use mvnw to lock the version.


Eskibear:
@Flanker32 This project fails to build with JDK 10+ . Here I simply set the ubuntu dist to `trusty` which uses JDK8

PR #202, opened 2019-11-12
Use workspace folder name as the invisible project name

jdneo:
Need some feedback:

WeChat Screenshot_20191112145139


Vigilans:
Seems apparently that '_' needs another name.

Here's a comparison with invisible project and a project imported to Eclipse IDE:

image

Vigilans:
I was thinking whether an opened vscode folder will contain multiple invisible projects...

Now I think multiple projects should be delegated to vscode workspace mechanism -- one invisible project per workspace folder.

But seems that for now invisible project does not support workspace so well?
image
image

jdneo:
Hopefully this is what we would achieve:
image

I'll update the PR recently.

jdneo:
Per offline discussion, the fix about skipping empty entry will be provided by @Vigilans in another PR.

jdneo:
Per offline discussion, reserve the internal project name in the explorer to provide transparency

Issue #180, opened 2019-09-23
Please consider to open project folder in current window rather than to launch a new one


byyxx128:
**This may be actually not a bug, but it is definitely unfriend for users.**
Ref: https://code.visualstudio.com/docs/java/java-project
1. Press Ctrl+Shift+P to open Command Palette.
2. Type `java` and click `Java: Create Java Project`.
3. Choose a location.
4. Input project name and press `Enter` on keyboard.

Then a new window of VS Code will be launched.

Everything seems good till now. But why does it always open a new window?
How about to open the project folder in current window directly?

PR #204, opened 2019-11-12
Open in current window when no workspace folder is available

jdneo:
fix https://github.com/microsoft/vscode-java-dependency/issues/180

### DIFF
--- a/src/controllers/projectController.ts
+++ b/src/controllers/projectController.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 import * as fse from "fs-extra";
+import * as _ from "lodash";
 import * as path from "path";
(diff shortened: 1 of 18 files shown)


src/controllers/projectController.ts
```diff
@@ -43,7 +44,10 @@ export class ProjectController {
             return;
         }
         if (await this.scaffoldJavaProject(basePath, projectName, javaVersion)) {
-            return commands.executeCommand("vscode.openFolder", Uri.file(path.join(basePath, projectName)), true);
+            if (!workspace || _.isEmpty(workspace.workspaceFolders)) {
```
akaroml: const openInNewWindow = !workspace || _.isEmpty(workspace.workspaceFolders);
return commands. executeCommand(......, openInNewWindow);

Issue #207, opened 2019-11-14
no right click action on file or folder


jason919:
the flat package view is nice, however, I lose all the right click functions on files and folders.
Not sure if that's even possible with current VS code

testforstephen:
We don't support right click on the **Java Dependency View** yet. Do you mean you lose the right click on the **File Explorer**?

jason919:
Right click still works on the file explorer. 
If the java dependency view supports all the right click, then it can replace the file explorer

maxandersen:
+1 on this as without it the java dependency view is not really adding much value besides being a nice indicator that Java LSP actually is enabled. For now I'm minimizing the java dependency view as all the contextual actions I seem to use all the time are missing.

testforstephen:
Menus in progress:
- [x] Reveal in Explorer
- [x] Copy Path
- [x] Copy Relative Path
- [ ] Copy Qualified Name (enable for package and class)

If you want more menus to be added to dependency view, please add comment below.

maxandersen:
How about all of them ? :)

Here are the ones I keep going for and then have to go to resource view on:

- [ ] Open to the Side
- [ ] Reveal in Finder
- [ ] Open in Terminal (for directories mainly)
- [ ] all the revision commands
- [ ] cut, copy, rename, delete

0dinD:
Any updates on this? I noticed [you can now](https://github.com/microsoft/vscode-java-dependency/pull/299) create new classes/packages from the context menu, which is great! But it's really annoying to have to go back to the file explorer when you need to rename or delete said class/package. I agree with the previous comment, but rename and delete on class/package nodes is the feature that I miss the most!

jdneo:
@0dinD Thank you for the suggestion. That's a valid request. I'll take a look these days.

benzman81:
- [ ] New File
- [ ] New Folder

see #598

jdneo:
New File and Folders are implemented in https://github.com/microsoft/vscode-java-dependency/pull/742.

For other actions, there are separate issues to track:
- https://github.com/microsoft/vscode-java-dependency/issues/743
- https://github.com/microsoft/vscode-java-dependency/issues/630

I'm closing this issue. Any new request please file issues separately for each of the actions. Thanks.

PR #208, opened 2019-11-19
Support right click context for viewer nodes

Vigilans:
It contributes menus such as **Reveal in Explorer**, **Copy Path** for #207

Based on:
- [x] #215

Demo:
image


package.nls.json
```diff
@@ -7,6 +7,9 @@
   "contributes.commands.java.view.package.changeToHierarchicalPackageView":"Change to hierarchical package representation",
   "contributes.commands.java.view.package.linkWithFolderExplorer":"Synchronize dependency viewer selection with folder explorer",
   "contributes.commands.java.view.package.unlinkWithFolderExplorer":"Desynchronize dependency viewer selection with folder explorer",
+  "contributes.commands.java.view.package.revealFileInOS": "Open Containing Folder",
```
testforstephen: => Reveal in Explorer

Vigilans: Description updated

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java
```diff
@@ -161,17 +162,29 @@ public static PackageNode createNodeForClasspathEntry(IClasspathEntry classpathE
...
             }
             if (container != null) {
-                switch (nodeKind) {
-                case CONTAINER:
-                    return new ContainerNode(container.getDescription(), container.getPath().toPortableString(), nodeKind, entry.getEntryKind());
-                case PACKAGEROOT:
+                PackageNode node = null;
+                if (nodeKind == NodeKind.CONTAINER) {
+                    node = new ContainerNode(container.getDescription(), container.getPath().toPortableString(), nodeKind, entry.getEntryKind());
+                    switch (container.getKind()) {
```
testforstephen: extract to a method.

package.json
```diff
@@ -83,6 +83,21 @@
           "dark": "images/dark/icon-link.svg",
           "light": "images/light/icon-link.svg"
         }
+      },
+      {
+        "command": "java.view.package.revealFileInOS",
```
testforstephen: revealFileInOS -> revealInExplorer

Vigilans: `revealInExplorer` is not appropriate since it is related to two built-in commands in VS Code: `revealInExplorer` and `git.revealInExplorer`: They all refer to the meaning of "Reveal the file node in VS Code File Explorer", not the OS explorer.

In contrast, `revealFileInOS` is a built-in command of VS Code to open a file in OS explorer, and our command just forwards the dependency node's uri to it, so this naming would be fine.

testforstephen: so why not to mount the VS Code command to the dependency explorer directly?

Vigilans: It would be best that we could mount the builtin command directly. 
But the builtin command accepts a `Uri` argument, where context entry specified in `package.json` only allows for passing a `ExplorerNode` to the command binding.

(The `TreeItem.command` property can indeed specify the argument to be passed, but this is not the case with `view/item/context` entry)

testforstephen: It seems VS Code provides different names in different OSes, could you check whether we need keep consistent?

Vigilans: Sorry for I didn't notice this comment...

As [this file](https://github.com/microsoft/vscode/blob/0dbe568b59c9fd213bddc5ddc730a4892f3b011c/src/vs/workbench/contrib/files/electron-browser/fileActions.contribution.ts#L64-L67) shows, VS Code does provide different names here:
* Windows: `Reveal In Explorer`
* Mac OS: `Reveal In Finder`
* Linux: `Open Containing Folder`

And the [Chinese translation](https://github.com/microsoft/vscode-loc/blob/257d355e31d868801fb36be24c6070a3c39885a7/i18n/vscode-language-pack-zh-hans/translations/main.i18n.json#L3905-L3910) uses that localization:
```json
"vs/workbench/contrib/files/electron-browser/fileActions.contribution": {
    "revealInWindows": "在资源管理器中显示",
    "revealInMac": "在 Finder 中显示",
    "openContainer": "打开所在的文件夹",
    "filesCategory": "文件"
},
```

Currently I have no knowledge of how to utilize the localization here according to OS...

testforstephen: Currently i'm OK to keep the same one, if there is user complaint, then fix it.
A workaround is add three difference commands, and use when clause to enable the correct one.

PR #210, opened 2019-11-19
Do not display workspace link '_' for invisible project

Vigilans:
Before:

After:
image

--- Updated---
When root has content
_+src

When root is empty


Eskibear (approved):

jdneo:
Thanks @vigilans

PR #212, opened 2019-11-19
Refactor onDidChangeConfiguration handler in favor of listener pattern

Vigilans:
This PR
* Refactor the `onDidChangeConfiguration` hanlder in favor of`listener(updatedConfig, oldConfig)` pattern.
* move the old logic into separate listeners
* makes sure that when a listener gets called, it will receive the **latest** configuration from `Settings.getSomeSetting`.

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -17,10 +17,9 @@ export class Settings {
                 return;
             }
             const oldConfig = this._dependencyConfig;
-            const updatedConfig = workspace.getConfiguration("java.dependency");
-            this._dependencyConfig = updatedConfig;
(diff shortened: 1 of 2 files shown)


src/settings.ts
```diff
@@ -16,23 +16,27 @@ export class Settings {
             if (!e.affectsConfiguration("java.dependency")) {
                 return;
             }
+            const oldConfig = this._dependencyConfig;
             const updatedConfig = workspace.getConfiguration("java.dependency");
```
testforstephen: updatedConfig variable is not necessary.

PR #213, opened 2019-11-19
Support managing referenced libraries

### LINKED ISSUE #174: Can you add eclipse user_library support in the next version?
## Suggestion

Can you add eclipse user_library support in the next version?

## Use Cases

Every time I add a lot of jars into my project, it will be very troublesome,
So I need an easier way to add jars.

## Examples
Like this

    <classpathentry kind="con" path="org.eclipse.jdt.USER_LIBRARY/7.0sp3"/>

But in the current version it doesn't work.

Vigilans:
Fix #174, https://github.com/microsoft/vscode-java-pack/issues/94 

Based on:
- [x] https://github.com/eclipse/eclipse.jdt.ls/pull/1267
- [x] #212
- [x] #215 

VS Code will take care of the file watching and searching work by providng a setting:

   ```typescript
   "java.dependency.referencedLibraries": string[] | {
     "include": string[],
     "exclude": string[],
     "sources": { [binary: string]: string }
   }
   ```

   ...with the following extended behaviors before sending to `jdt.ls`:

   * If specified setting is an array, it will be expanded into a full object, with the original content as `include` field. 
   * Jars in `lib/` but not `included` will be appended to `exclude` field. It is to ensure jars not listened by VS Code don't get shown in referenced libraries — only one source of truth. (This is not an elegent workaround, waiting for better alternative...)
   * A match of `foo.jar` and `foo-src.jar` (not necessarily in the same folder) will be added to sources mapping.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -1,6 +1,8 @@
 {
   "description": "Manage Java Dependencies in VSCode",
   "contributes.commands.java.project.create": "Create Java Project",
+  "contributes.commands.java.project.addLibraries": "Add a jar file or a folder to project library reference",
+  "contributes.commands.java.project.removeLibrary": "Remove jar file from project library reference",
   "contributes.commands.java.view.package.refresh": "Refresh",
   "contributes.commands.java.view.package.changeRepresentation": "Change package representation",
   "contributes.commands.java.view.package.changeToFlatPackageView":"Change to flat package representation",
@@ -14,6 +16,6 @@
   "configuration.java.dependency.showOutline": "Enable show outline in the Java Dependency explorer",
   "configuration.java.dependency.syncWithFolderExplorer": "Synchronize dependency viewer selection with folder explorer",
   "configuration.java.dependency.autoRefresh": "Synchronize dependency viewer with changes",
-  "configuration.java.dependency.refreshDelay": "The delay time (ms) the auto refresh is invoked when changes are detected.",
+  "configuration.java.dependency.refreshDelay": "The delay time (ms) the auto refresh is invoked when changes are detected",
   "configuration.java.dependency.packagePresentation": "Package presentation mode: flat or hierarchical"
 }
--- a/package.nls.json
+++ b/package.nls.json
@@ -1,8 +1,8 @@
 {
   "description": "Manage Java Dependencies in VSCode",
   "contributes.commands.java.project.create": "Create Java Project",
-  "contributes.commands.java.project.addLibraries": "Add a jar file or a folder to project library reference",
-  "contributes.commands.java.project.removeLibrary": "Remove jar file from project library reference",
+  "contributes.commands.java.project.addLibraries": "Add a jar file or a folder to project classpath",
+  "contributes.commands.java.project.removeLibrary": "Remove jar file from project classpath",
   "contributes.commands.java.view.package.refresh": "Refresh",
   "contributes.commands.java.view.package.changeRepresentation": "Change package representation",
   "contributes.commands.java.view.package.changeToFlatPackageView":"Change to flat package representation",
--- a/src/controllers/libraryController.ts
+++ b/src/controllers/libraryController.ts
@@ -1,0 +1,74 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as chokidar from "chokidar";
+import * as fse from "fs-extra";
+import * as _ from "lodash";
+import { commands, Disposable, ExtensionContext, Uri, window, workspace, WorkspaceFolder } from "vscode";
+import { instrumentOperation } from "vscode-extension-telemetry-wrapper";
+import { Commands } from "../commands";
+import { Jdtls } from "../java/jdtls";
+import { Settings } from "../settings";
+import { Utility } from "../utility";
+import { DataNode } from "../views/dataNode";
+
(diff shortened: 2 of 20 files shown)


testforstephen:
what's the status for this PR?

Vigilans:
Some important modifications are saved in my Linux PC. May I rebase to the master branch and update the PR after going back home tonight?

src/controllers/libraryController.ts
```diff
@@ -0,0 +1,74 @@
...
+import { instrumentOperation } from "vscode-extension-telemetry-wrapper";
+import { Commands } from "../commands";
+import { Jdtls } from "../java/jdtls";
+import { Settings } from "../settings";
+import { Utility } from "../utility";
+import { DataNode } from "../views/dataNode";
+
+export class LibraryController implements Disposable {
+
+    private libraryWatcher: chokidar.FSWatcher | undefined = undefined;
```
testforstephen: I don't see anyone is using this watcher. Should remove it?

Vigilans: Oh, I forgot to remove this line.

Vigilans: I've update the PR. Apart from removing the watcher, I also refactor the controllers to dispose the commands by itself, rather than pushing it to extension context.

src/controllers/libraryController.ts
```diff
@@ -0,0 +1,72 @@
...
+            if (!results) {
+                return;
+            }
+            libraryGlobs = await Promise.all(results.map(async (uri: Uri) => {
+                const uriPath = workspace.asRelativePath(uri, false);
+                return (await fse.stat(uri.fsPath)).isDirectory() ? `${uriPath}/**/*.jar` : uriPath;
+            }));
+        }
+        const setting = Settings.referencedLibraries();
+        if (Settings.isDefaultReferencedLibraries()) {
```
testforstephen: Why need check the default value?
why not use `setting.include = setting.include || []`?

webpack.config.js
```diff
@@ -34,7 +34,10 @@ const config = {
             use: [{
                 loader: 'ts-loader',
             }]
+        }, {
+            test: /\.node$/,
+            loader: 'native-ext-loader',
```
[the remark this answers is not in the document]
Vigilans: It is the legacy of introducing `chokidar` --> `fsevent`, where a `fsevent.node` is introduced and webpack needs this loader for bundling it. 

Leaving this loader may help the introduction of packages containing a native node extension like `fsevent`, but removing it in this PR is just fine.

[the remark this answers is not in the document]
testforstephen: Please remove it. Should not add unnecessary code for some imaginary features in future.

src/controllers/libraryController.ts
Vigilans: The `isDefaultReferencedLibraries` checks whether the `referencedLibraries` setting has been modified by user, instead of checking whether `setting.include` exists. In fact, `setting.include` always exists.

The motivation here is when a user adds a library at the first time, the default value `lib/**/*.jar` should not go with the added libraries. e.g., when adding `data/**/*.jar`, the setting in the workspace should be like this:
```json
"java.project.referencedLibraries": [
    "data/**/*.jar"
]
```
But without this check, it will be like this:
```json
"java.project.referencedLibraries": [
    "lib/**/*.jar",
    "data/**/*.jar"
]
```
Which will make the user confusing.

package.nls.json
```diff
@@ -1,6 +1,8 @@
 {
   "description": "Manage Java Dependencies in VSCode",
   "contributes.commands.java.project.create": "Create Java Project",
+  "contributes.commands.java.project.addLibraries": "Add a jar file or a folder to project library reference",
```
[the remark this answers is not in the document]
Vigilans: Is it suitable to introduce the concept of `classpath` to the user using invisible project?

src/controllers/libraryController.ts
Vigilans: But maybe this is also a confusing behavior. Because removing the "lib/**/*.jar" may cause some jar dependency lost. I will remove this line.

package.nls.json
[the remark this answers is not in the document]
testforstephen: classpath is a standard concept from JDK, not eclipse special. I think it's OK.

src/controllers/libraryController.ts
```diff
@@ -0,0 +1,69 @@
...
+            if (!results) {
+                return;
+            }
+            libraryGlobs = await Promise.all(results.map(async (uri: Uri) => {
+                const uriPath = workspace.asRelativePath(uri, false);
+                return (await fse.stat(uri.fsPath)).isDirectory() ? `${uriPath}/**/*.jar` : uriPath;
+            }));
+        }
+        const setting = Settings.referencedLibraries();
+        setting.include.push(...libraryGlobs);
```
jdneo: Simply push might not work in some scenario, like:

if the original setting is:

```json
"java.project.referencedLibraries": {
        "include": [
            "lib/**/*.jar"
        ],
        "exclude": [
            "lib/junit-platform-console-standalone-1.5.2.jar"
        ]
    }
```

where the `junit-platform-console-standalone-1.5.2.jar` is the only jar in the lib folder.

Now if I re-add the lib folder into the referenced libs, the exclude part won't be updated.

PR #220, opened 2020-01-09
bugfix: No response when open a file with jdt uri

### LINKED ISSUE #219: Right click on folder under jar doesn't work
#### Configuration:
```
VScode: 1.41.1
JDK: 1.8.0_231
OS: Win
Java Dependency view version: 0.7.0
Language Support for Java version:0.55.1
```

#### Repro Steps:
1. Open a project in vscode
2. Open java dependecies viewer, 
3. Right click the file under folder and click "reveal in explorer" 
image

#### Result:
Right click on a folder under .jar, it doesn't work.
Right click on a file under the folder in .jar, "Reveal in Explorer" doesn't work

jdneo:
fix #219

### DIFF
--- a/src/views/dataNode.ts
+++ b/src/views/dataNode.ts
@@ -58,7 +58,7 @@ export abstract class DataNode extends ExplorerNode {
 
     protected computeContextValue(): string {
         let contextValue = this.contextValue;
-        if (this.uri && Uri.parse(this.uri).scheme === "file") {
+        if (this.uri && this.uri.startsWith("file")) {
             contextValue = `${contextValue || ""}+uri`;
         }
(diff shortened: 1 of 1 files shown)


src/views/dataNode.ts
```diff
@@ -58,7 +58,7 @@ export abstract class DataNode extends ExplorerNode {
 
     protected computeContextValue(): string {
         let contextValue = this.contextValue;
-        if (this.uri) {
+        if (this.uri && this.uri.startsWith("file")) {
```
testforstephen: startsWith("file:")?

jdneo: Yes that's better

PR #223, opened 2020-01-14
bugfix: Cannot remove jar if it's already in the include field

### LINKED ISSUE #222: The remove button does not work when the include and exclude has the same artifact
demo

jdneo:
fix #222

### DIFF
--- a/src/controllers/libraryController.ts
+++ b/src/controllers/libraryController.ts
@@ -64,7 +64,13 @@ export class LibraryController implements Disposable {
 
     public async removeLibrary(library: string) {
         const setting = Settings.referencedLibraries();
-        setting.exclude = this.updatePatternArray(setting.exclude, workspace.asRelativePath(library, false));
+        const removedPaths = _.remove(setting.include, (include) => {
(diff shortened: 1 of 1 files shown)


src/controllers/libraryController.ts
```diff
@@ -62,9 +63,19 @@ export class LibraryController implements Disposable {
...
 
-    public async removeLibrary(library: string) {
+    public async removeLibrary(removalFsPath: string) {
         const setting = Settings.referencedLibraries();
-        setting.exclude = this.updatePatternArray(setting.exclude, workspace.asRelativePath(library, false));
+        const removedPaths = _.remove(setting.include, (include) => {
+            if (path.isAbsolute(include)) {
+                return Uri.file(include).fsPath === removalFsPath;
+            } else {
+                return include === workspace.asRelativePath(removalFsPath, false);
```
Vigilans: Also apply `asRelativePath` on include to ensure they're in the same form?
```suggestion
                return workspace.asRelativePath(include, false) === workspace.asRelativePath(removalFsPath, false);
```

PR #236, opened 2020-02-14
Prepare for 0.9.0 release

jdneo:


.vscode/settings.json
```diff
@@ -1,2 +1,3 @@
 {
+    "java.configuration.updateBuildConfiguration": "automatic"
```
testforstephen: commit by mistake?

package.json
```diff
@@ -226,22 +226,22 @@
         },
         {
           "command": "java.project.addLibraries",
-          "when": "view == javaDependencyExplorer && viewItem =~ /java:container\/referenced-libraries$/",
+          "when": "view == javaDependencyExplorer && viewItem =~ /java:container/referenced-libraries$/",
```
testforstephen: looks like this is a regular pattern. Is there any reason to remove the escape char `\/` ?

jdneo: My bad, `npm install` automatically removes them

PR #252, opened 2020-04-13
Refine the command to create Java project

### LINKED ISSUE #199: Creating invisible project instead of Eclipse project
Benefit we can get from this change:
- Support different kind of JDK automatically
- Much easier to configure the dependencies.

### LINKED ISSUE #206: Get the Java runtime information from Java Language Server
We host a lot of logic by our own, which should be get rid of: https://github.com/microsoft/vscode-java-dependency/blob/master/src/utility.ts#L14-L62

### LINKED ISSUE #249: Support creation of Maven Projects
Currently, the feature "Create Maven Project" is part of the Maven extension.

It is confusing to Java developers, once they have both extensions installed, to have two different "Create Project" actions in the command pallet.

I suggest that this dependency becomes responsible for creating projects (Eclipse, Maven, and Gradle).

Also, see #248

jdneo:
fix #249
fix #206
fix #199


jdneo:
@Eskibear @testforstephen @akaroml 

Please take a look when you have time. The A/B testing work will be addressed in another PR, to avoid too much code change in one PR.

src/controllers/projectController.ts
```diff
@@ -25,10 +26,36 @@ export class ProjectController implements Disposable {
...
-        if (!javaVersion) {
-            return;
+        const projectKinds: QuickPickItem[] = [{
+            label: BuildTool.None,
+            detail: "A project without any build tools",
+        }];
+        if (contextManager.getContextValue(Context.MAVEN_ENABLED)) {
+            projectKinds.push({
+                label: BuildTool.Maven,
+                detail: "Use Maven as the build tool of your Java project",
```
Eskibear, line 36: > Apache Maven is a software project management and comprehension tool. Based on the concept of a project object model (POM), Maven can manage a project's build, reporting and documentation from a central piece of information.

so would something like "Use Maven to manage your project" be simple and concise?

src/controllers/projectController.ts
```diff
@@ -25,10 +26,36 @@ export class ProjectController implements Disposable {
...
+            });
+        }
+        const choice: QuickPickItem | undefined = projectKinds.length === 1 ? projectKinds[0] :
+            await window.showQuickPick(projectKinds, {
+                ignoreFocusOut: true,
+                placeHolder: "Select the project build tool",
+            },
+        );
+
+        switch (choice.label) {
```
Eskibear, line 46: `choice` is possible to be `undefined`, if you have more than one project kinds.

PR #260, opened 2020-05-11
Integrate the ExP framework

jdneo:


package.json
```diff
@@ -267,29 +267,30 @@
...
+    "webpack": "^4.42.0",
     "webpack-cli": "^3.3.11"
   },
   "dependencies": {
     "fs-extra": "^7.0.1",
     "lodash": "^4.17.15",
     "minimatch": "^3.0.4",
-    "vscode-extension-telemetry-wrapper": "^0.7.1"
+    "vscode-extension-telemetry-wrapper": "^0.7.1",
+    "vscode-tas-client": "0.0.702"
```
daytonellwanger, line 294: You should actually use version `0.0.659` of `vscode-tas-client` for now. `0.0.702` has a bug that I'll resolve later this week.

src/ExpManager.ts
```diff
@@ -0,0 +1,40 @@
...
+    public initialize(exp: IExperimentationService | undefined) {
+        this.exp = exp;
+    }
+
+    public async isFlightEnabled(flight: string): Promise<boolean> {
+        if (!this.exp) {
+            return false;
+        }
+
+        try {
```
daytonellwanger, line 19: The `vscode-tas-client` package will query the treatment assignment service in the background (every 30 minutes), so flight information should never be much staler than the last opened instance of VS Code. If you want to be absolutely sure your extension has up-to-date flight information, you can do this. However, it probably isn't necessary.

src/ExperimentationService.ts
```diff
@@ -0,0 +1,35 @@
...
+
+let expService: IExperimentationService;
+
+export function getExpService() {
+    return expService;
+}
+
+export function init(context: vscode.ExtensionContext) {
+    const extensionName = "vscjava.vscode-java-dependency";
+    const extensionVersion = "0.10.1";
```
daytonellwanger, line 32: It might be better to read this from your `package.json` so you don't have to remember to update this when you change the extension version.

src/controllers/projectController.ts
```diff
@@ -31,10 +32,16 @@ export class ProjectController implements Disposable {
             detail: "A project without any build tools",
         }];
         if (contextManager.getContextValue(Context.MAVEN_ENABLED)) {
-            projectKinds.push({
+            const isMavenDefault: boolean = await expManager.isFlightEnabled("defaultMaven");
```
daytonellwanger, line 35: Are you OK paying the async cost here? Given that this could result in an http request, this could be 1-5 seconds.

src/ExpManager.ts
jdneo: So does that mean that we can just call `isCachedFlightEnabled()` here, and call `isFlightEnabledAsync()` for 1 time in the `init` method?

daytonellwanger: You shouldn't ever need to call `isFlightEnabledAsync` unless you want the absolute latest flight information. But I can't think of any scenarios where you would.

jdneo: I see. Thank you @daytonellwanger.

Another question, is below code needed in the init functions?

```ts
const asyncDummyCheck = (arg: any) => {
        expService?.isFlightEnabledAsync('dummy').then((v) => { return; }).catch((r) => { return; });
    };
    expService?.isCachedFlightEnabled('dummy').then(asyncDummyCheck).catch(asyncDummyCheck);
}
```

daytonellwanger: This should no longer be needed with the version of the `vscode-tas-client` package I mentioned above.

jdneo: I see. Thank you for your help @daytonellwanger. :)

I'll update the code according to the comments.

src/ExperimentationService.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+let expService: IExperimentationService;
+
+export function getExpService() {
+    return expService;
+}
+
+export async function init(context: vscode.ExtensionContext): Promise<void> {
+    const packageJson: {} = await fse.readJSON(context.asAbsolutePath("package.json"));
```
daytonellwanger, line 32: You might be able to just do `require('package.json')`?

src/ExperimentationService.ts
```diff
@@ -0,0 +1,46 @@
...
+    // tslint:disable: no-string-literal
+    const extensionName = `${packageJson["publisher"]}.${packageJson["name"]}`;
+    const extensionVersion = packageJson["version"];
+    // tslint:enable: no-string-literal
+    expService = getExperimentationService(extensionName, extensionVersion,
+        TargetPopulation.Public, new ExperimentationTelemetry(), context.globalState);
+
+    // Due to a bug in the tas-client module, a call to isFlightEnabledAsync is required to begin
```
daytonellwanger, line 40: You should be able to drop these dummy checks

jdneo: I tried to remove these lines. But I found that when I'm calling `getExpService().isCachedFlightEnabled("defaultMaven");`, the cached features are empty.

image

Within the dummy code, the cached features are correct.

_// The vscode-tas-client I'm using is 0.0.659_

daytonellwanger: Oh interesting. Maybe we still have another bug to track down. OK, keep these checks for now and I'll look into the issue ASAP, and with the next update to the package, you can remove them.

jdneo:
Hey @daytonellwanger，

I have updated the PR. Please take a look when you have time. Thanks a lot.

src/controllers/projectController.ts
```diff
@@ -31,10 +32,16 @@ export class ProjectController implements Disposable {
             detail: "A project without any build tools",
         }];
         if (contextManager.getContextValue(Context.MAVEN_ENABLED)) {
-            projectKinds.push({
+            const isMavenDefault: boolean = await getExpService().isCachedFlightEnabled("defaultMaven");
```
daytonellwanger, line 35: null check after `getExpService`?

PR #299, opened 2020-08-13
Support new class & package from explorer

### LINKED ISSUE #78: Create new class and package from explorer
Is it possible to add a context menu for functions such as `new class` and `new package`?

### LINKED ISSUE #126: The empty package can't display
**### Configuration:**
```
Vscode: 1.29.1
JDK: 1.8.0_191
OS: Win/Mac
Build: 60
```
The empty package can't display in the dependency view, as below:

image

jdneo:
resolve #78

Also fix #126

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
@@ -440,6 +440,8 @@ public class PackageCommand {
                 result.add(child);
             } else if (fragment.getNonJavaResources().length > 0) { // some package has non-java files
                 result.add(fragment);
+            } else if (!fragment.hasSubpackages()) {
+                result.add(fragment);
             }
         }
         Object[] nonJavaResources = root.getNonJavaResources();
--- a/src/explorerCommands/new.ts
+++ b/src/explorerCommands/new.ts
@@ -40,8 +40,9 @@ export async function newJavaClass(node: DataNode): Promise<void> {
 }
 
 function getNewFilePath(basePath: string, className: string): string {
-    const indexOfJavaExt: number = className.lastIndexOf(".java");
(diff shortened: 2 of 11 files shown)


src/explorerCommands/new.ts
```diff
@@ -0,0 +1,108 @@
...
+    // `workspace.applyEdit()` will trigger a workspace file event, and let the
+    // vscode-java extension to handle the type: class, interface or enum.
+    const workspaceEdit: WorkspaceEdit = new WorkspaceEdit();
+    const fsPath: string = getNewFilePath(packageFsPath, className);
+    workspaceEdit.createFile(Uri.file(fsPath));
+    workspace.applyEdit(workspaceEdit);
+}
+
+function getNewFilePath(basePath: string, className: string): string {
+    const indexOfJavaExt: number = className.lastIndexOf(".java");
```
testforstephen, line 43: One corner case is "org.java.MyClass", looks like you only need check whether className.endsWith(".java")

jdneo: yes, make sense

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
```diff
@@ -440,6 +440,8 @@
                 result.add(child);
             } else if (fragment.getNonJavaResources().length > 0) { // some package has non-java files
                 result.add(fragment);
+            } else if (!fragment.hasSubpackages()) {
+                result.add(fragment);
             }
```
testforstephen, line 445: looks like you could add the child directly `result.add(child)`. No need of check the conditions.

jdneo: That will make every sub packages get rendered in explorer (including the default package) but without any content. For all the empty packages, just show the leaf ones looks much better.

Screen Shot 2020-08-14 at 10 57 59 AM

PR #312, opened 2020-08-24
Use folder name as the name of the invisible project

jdneo:
image


jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java
```diff
@@ -275,6 +280,14 @@ public String getName() {
         return name;
     }
 
+    public String getDisplayName() {
+        return displayName;
```
testforstephen, line 284: return displayName || name;

testforstephen (approved):

PR #324, opened 2020-09-18
Fix backbutton issue

CsCherrYY:
Use an array to record the steps to fix the backbutton issue.
Also move `IStepMetadata` to a single file and do a small refactor in `IJarQuickPickItem` to avoid any misunderstanding.

### DIFF
--- a/src/exportJarSteps/IStepMetadata.ts
+++ b/src/exportJarSteps/IStepMetadata.ts
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
+
 import { Uri } from "vscode";
 import { ExportJarStep } from "../exportJarFileCommand";
 import { INodeData } from "../java/nodeData";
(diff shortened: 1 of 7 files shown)


src/exportJarSteps/IStepMetadata.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+import { Uri } from "vscode";
```
jdneo, line 3: Add an empty line

src/exportJarSteps/utility.ts
```diff
@@ -1,10 +1,11 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-import { QuickInputButtons, QuickPick, QuickPickItem, window } from "vscode";
+import { QuickInputButtons, QuickPick, QuickPickItem, Uri, window } from "vscode";
 
 export interface IJarQuickPickItem extends QuickPickItem {
```
jdneo, line 6: Move to each step according to the needs

jdneo (approved):

PR #331, opened 2020-09-22
feat: Add inline button for create new class

jdneo:
Add the `create new class` action as an inline button for `project`, `packageRoot` & `package` nodes

demo

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,8 @@
       {
         "command": "java.view.package.newJavaClass",
         "title": "%contributes.commands.java.view.package.newJavaClass%",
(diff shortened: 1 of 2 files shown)


Eskibear (approved):
LGTM, just wonder if we should use the same icon (+) with "create a project".  Consider icon `new-file` ?

jdneo:
> LGTM, just wonder if we should use the same icon (+) with "create a project". Consider icon `new-file` ?

That's a valid concern. @akaroml What do you think? IMO `new file` would be better.

testforstephen (approved):

PR #343, opened 2020-09-29
feat: Add 'delete' into the view context menu

xqzlgy2:
Enable User to delete package/file in context menu like below.
- All deleted file will be moved to Recycle Bin/Trash by default.
- Added `File/Type` into `ContextValueType` to distinguish common files and Java source files.
- Added `+enum/+class/+interface` as attributes to Type for detecting different kinds of source files.
- Added key binding for delete.
- Preserve focus when open file through Java project explorer

image
image

part of #207


0dinD:
Awesome, this is just what I wanted! If it's not too much to ask for, do you think you could add a 'rename' option to the menu? I think you can do it via the extension API, and hopefully vscode-java would be able to pick up on that rename event, in order to provide a refactoring to change the class name. (Which it does when you rename a java file from the file explorer.)

jdneo:
@0dinD Yes, `rename` will be added after this PR.

// BTW, we will have the National Holiday at our side so it will happen after the holiday (about in Mid. Oct.)

package.json
```diff
@@ -364,6 +373,21 @@
           "when": "view == javaProjectExplorer && viewItem =~ /java:project(?=.*?\\b\\+java\\b)(?=.*?\\b\\+uri\\b)/",
           "group": "new@20"
         },
+        {
+          "command": "java.view.package.moveFileToTrash",
+          "when": "view == javaProjectExplorer && viewItem =~ /java:(package|packageRoot)(?=.*?\\b\\+source\\b)(?=.*?\\b\\+uri\\b)/",
+          "group": "delete@10"
```
jdneo, line 379: Since we are having more and more items in the menu, I think this could be a good chance to setup some rules to group different commands in the menu.

Regarding to this, I think we can learn from what VS Code has done for the `File Explorer`, I suggest we group the delete command into `7_modification`. See: 
- https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups
- https://github.com/microsoft/vscode/blob/70e95ce5b9407c0df82d2c272f9730db1c307350/src/vs/workbench/contrib/files/browser/fileActions.contribution.ts#L540-L553

jdneo (changes requested):
@xqzlgy2

Could you please also investigate if it is possible to set default keyboard shortcut for this command in `package.json`?

package.json
xqzlgy2: I'll have a try to see if there is a good way to implement.

src/explorerCommands/delete.ts
```diff
@@ -0,0 +1,48 @@
...
+    const isFolder = children && children.length !== 0;
+    const message = getInformationMessage(node.name, isFolder);
+
+    const answer: string | undefined = await window.showInformationMessage(
+        message,
+        { modal: true },
+        confirmMessage,
+    );
+
+    if (!answer) {
```
jdneo, line 38: This if-block can be removed

src/explorerCommands/utils.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { DataNode } from "../views/dataNode";
+
+export function shouldModifyNode(node: DataNode): boolean {
```
jdneo, line 6: Rename to `isMutable()`

src/explorerCommands/delete.ts
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the MIT license.
+
+import { Uri, window, workspace } from "vscode";
+import { DataNode } from "../views/dataNode";
+import { ExplorerNode } from "../views/explorerNode";
+import { shouldModifyNode } from "./utils";
+
+const confirmMessage = "Move to Recycle Bin";
+
+function getInformationMessage(name: string, isFolder: boolean): string {
```
jdneo, line 11: nit: move the function to the bottom of this file

jdneo (changes requested):

src/explorerCommands/delete.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    const answer: string | undefined = await window.showInformationMessage(
+        message,
+        { modal: true },
+        confirmMessage,
+    );
+
+    if (answer === confirmMessage) {
+        const fileUri = Uri.parse(node.uri);
+        commands.executeCommand(Commands.CLOSE_OTHER_EDITORS_IN_GROUP, fileUri);
```
jdneo, line 34: try `await commands.executeCommand("workbench.action.closeEditorsInGroup", fileUri);`

jdneo (changes requested):

src/explorerCommands/delete.ts
```diff
@@ -0,0 +1,47 @@
...
+        workspace.fs.delete(fileUri, {
+            recursive: true,
+            useTrash: true,
+        });
+    }
+}
+
+function getInformationMessage(name: string, isFolder: boolean): string {
+    const folderMsg = isFolder ? " and its contents" : "";
+    const msg = `Are you sure you want to delete \'${name}\'${folderMsg}?\n`;
```
jdneo, line 44: How about add an empty line between the two sentence?

src/explorerCommands/delete.ts
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { commands, Uri, window, workspace } from "vscode";
+import { Commands } from "../commands";
```
jdneo, line 5: Remove the unused imports

jdneo (approved):

jdneo:
Thanks @xqzlgy2!

PR #350, opened 2020-10-15
Support custom task to export jar

CsCherrYY:
This PR enables project manager to support custom task to export jar. The available template of tasks can be found at `Terminal->Configure Tasks...`
The user can change the value of the fields to custom what to export. The following is the available fields:
```
"type": "java" // should not be edited.
"label": "java: exportjar:demo-noMain" // The label of this task to show in the wizard from `Terminal->Run Task...`
"mainClass": "demo.App", // The mainClass in `MANIFEST.MF`. If this field is not specified, the wizard selecting the mainClass will appear during the export.
"targetPath": "${workspaceFolder}/${workspaceFolderBasename}.jar", //The targetPath of this jar. Can use VS Code variables. If this field is not specified, the export process will apply the setting `Java->Project->Export Jar->TargetPath` in settings.json.
"elements": [ //The elements to export. If this field is not specified, the wizard selecting the elements will appear during the export.
	"${compileOutput}", //Reserved variable, which stands for the output path of this project.
	"${testCompileOutput}", //Reserved variable, which stands for the test output path of this project.
	"${dependencies}", //Reserved variable, which stands for the runtime libraries of this project.
	"${testDependencies}", //Reserved variable, which stands for the test libraries of this project.
	"!${compileOutput}/demo/innerPkg", //The glob pattern to specific which to exclude in the elements. Can also include reserved variables.
	"C:/work/demo/test.iml" //The glob pattern which supports both absolute path and relative path to the project folder.
],
```

The user can run the custom task at `Terminal->Run Task` and select the `java` type. The quickpick box will show all the available tasks.


src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -29,14 +46,74 @@ export class ExportJarTaskProvider implements TaskProvider {
         return task;
     }
 
+    private static exportJarTaskProvider: Disposable | undefined;
+
+    private tasks: Task[] | undefined;
+
     public async resolveTask(_task: Task): Promise<Task> {
```
jdneo, line 53: Do not use `_` prefix if the param is used.

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -29,14 +46,74 @@ export class ExportJarTaskProvider implements TaskProvider {
...
-        return _task;
+        const definition: IExportJarTaskDefinition = <IExportJarTaskDefinition>_task.definition;
+        const folder: WorkspaceFolder = <WorkspaceFolder>_task.scope;
+        const stepMetadata: IStepMetadata = {
+            entry: undefined,
+            workspaceFolder: folder,
+            steps: [],
+        };
+        const task: Task = new Task(definition, folder, _task.name, ExportJarTaskProvider.exportJarType,
```
jdneo, line 61: Will `definition`/`folder` be undefined?

src/exportJarSteps/utility.ts
```diff
@@ -8,6 +8,12 @@ import { ExportJarStep } from "../exportJarFileCommand";
 import { IStepMetadata } from "./IStepMetadata";
 
 export const SETTING_ASKUSER: string = "askUser";
+export const RUNTIME_DEPENDENCIES: string = "RuntimeDependencies";
```
jdneo, line 11: Group them into a namespace

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -1,18 +1,35 @@
...
+import { IClassPaths, IStepMetadata } from "./IStepMetadata";
+import { PathTrie } from "./PathTrie";
+import { COMPILE_OUTPUT, DEFAULT_OUTPUT_PATH, RUNTIME_DEPENDENCIES, SETTING_ASKUSER, TEST_DEPENDENCIES, TESTCOMPILE_OUTPUT } from "./utility";
 
 export class ExportJarTaskProvider implements TaskProvider {
 
     public static exportJarType: string = "java";
 
+    public static setProvider(): void {
+        ExportJarTaskProvider.exportJarTaskProvider = tasks.registerTaskProvider(ExportJarTaskProvider.exportJarType, new ExportJarTaskProvider());
```
jdneo, line 26: confusing method name. A setter should have params

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -29,14 +46,74 @@ export class ExportJarTaskProvider implements TaskProvider {
...
+            new CustomExecution(async (resolvedDefinition: TaskDefinition): Promise<Pseudoterminal> => {
+                return new ExportJarTaskTerminal(resolvedDefinition, stepMetadata);
+            }));
+        task.presentationOptions.reveal = TaskRevealKind.Never;
+        return task;
     }
 
     public async provideTasks(): Promise<Task[]> {
-        return [];
+        if (this.tasks !== undefined) {
```
jdneo, line 70: undefined or isEmpty?

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,13 +134,160 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+        for (const classpath of classpathResult.modulepaths) {
+            if (extname(classpath) === ".jar") {
+                dependencies.push(classpath);
+            } else {
+                classpaths.push(classpath);
+            }
+        }
+    }
+
+    private async resolveClassPaths(runtimeDependencies: string[], testDependencies: string[],
```
jdneo, line 198: Can we use a utility function? This method has very limited extensibility

src/exportJarSteps/PathTrie.ts
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jdneo, line 1: Can we reuse Trie?

src/exportJarSteps/GenerateJarExecutor.ts
```diff
@@ -91,22 +91,28 @@ export class GenerateJarExecutor implements IExportJarStepExecutor {
...
             });
         });
         if (_.isEmpty(dependencyItems)) {
             throw new Error("No classpath found. Please make sure your java project is valid.");
         } else if (dependencyItems.length === 1) {
-            stepMetadata.elements.push(dependencyItems[0].path);
+            const classpath: IClassPaths = {
+                source: dependencyItems[0].path,
+                destination: undefined,
+                isExtract: false,
```
jdneo, line 113: What's the meanning of `isExtract`

jdneo (changes requested):

src/exportJarSteps/ExportJarTaskProvider.ts
CsCherrYY: Neither. `definition` and `folder` come from the pre-defined task generated by `provideTasks()` both.

src/exportJarSteps/GenerateJarExecutor.ts
CsCherrYY: If we want to extract the sources rather than just copy them into the jar file, for example, the dependencies, we will set this field to true. Otherwise false.

jdneo: Ok, so is this a user visible field?

CsCherrYY: No. We will set this field to `true` if an `IClassPaths` was generated from dependencies. But I will also change this name to a more readable one in the next commit.

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
```diff
@@ -160,14 +167,16 @@ public static exportResult exportJar(List<Object> arguments, IProgressMonitor mo
...
-                    if (classpath.endsWith(".jar")) {
-                        ZipFile zip = new ZipFile(classpath);
-                        writeArchive(zip, true, true, target, directories, monitor);
+            for (ClassPath classpath : classpaths) {
+                if (classpath.isExtract) {
+                    ZipFile zip = new ZipFile(classpath.source);
+                    writeArchive(zip, true, true, target, directories, monitor);
+                } else {
+                    File file = new File(classpath.source);
+                    if (StringUtils.isEmpty(classpath.destination)) {
```
jdneo, line 176: why we need `writeFileRecursively` when `destination` is empty?

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -1,18 +1,36 @@
...
+import { IClasspathResult } from "./GenerateJarExecutor";
+import { IClassPath, IStepMetadata } from "./IStepMetadata";
+import { ExportJarProperties } from "./utility";
 
 export class ExportJarTaskProvider implements TaskProvider {
 
     public static exportJarType: string = "java";
 
+    public static registerProvider(): void {
+        ExportJarTaskProvider.exportJarTaskProvider = tasks.registerTaskProvider(ExportJarTaskProvider.exportJarType, new ExportJarTaskProvider());
```
jdneo, line 27: This is a disposable

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,13 +135,173 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+            if (!isAbsolute(positivePath)) {
+                return "!" + join(this.stepMetadata.workspaceFolder.uri.fsPath, positivePath);
+            }
+        } else if (!isAbsolute(path)) {
+            return join(this.stepMetadata.workspaceFolder.uri.fsPath, path);
+        }
+        return path;
+    }
+
+    private async getClasspaths(projectUri: string, _scope: string, classpaths: string[], dependencies: string[]): Promise<void> {
```
jdneo, line 179: if the variable is used, please do not add the prefix `_`

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,13 +135,173 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+        const extension: Extension<any> | undefined = extensions.getExtension("redhat.java");
+        const extensionApi: any = await extension?.activate();
```
jdneo, line 181: Do we have and guarantee to make sure it will not be undefined?

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,13 +135,173 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+    private async resolveClassPaths(runtimeDependencies: string[], testDependencies: string[],
+                                    classPathMap: Map<string, string[]>, testClassPathMap: Map<string, string[]>): Promise<IClassPath[]> {
+        // tslint:disable-next-line: no-invalid-template-strings
+        const regExp: RegExp = new RegExp("\\${(.*)}");
+        const classPathArray: string[] = [];
+        const dependencies: string[] = [];
+        for (const element of this.stepMetadata.elements) {
+            const variableResult = element.match(regExp);
+            if (_.isEmpty(variableResult) || variableResult.length <= 1) {
+                classPathArray.push(upath.normalizeSafe(this.toAbsolute(element)));
```
jdneo, line 208: Any reason not using `path.normalize`?

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,13 +135,173 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
         await createJarFile(this.stepMetadata);
         this.closeEmitter.fire();
     }
 
     public close(): void {
 
     }
+
+    private toAbsolute(path: string): string {
+        if (path.length > 0 && path[0] === "!") {
```
jdneo, line 168: 1. find the first index of `!`
2. if the index >= 0, set a boolean flag to true.
3. if the index >= 1, log a warning to tell the user that multiple `!` are not allowed and will be ignored.
4. extract a substring after the index
5. join if is not absolute
6. Add `!` if the boolean flag is true

src/exportJarSteps/GenerateJarExecutor.ts
```diff
@@ -138,7 +144,12 @@ export class GenerateJarExecutor implements IExportJarStepExecutor {
                             return;
                         }
                         for (const item of pickBox.selectedItems) {
-                            stepMetadata.elements.push(item.path);
+                            const classpath: IClassPath = {
+                                source: item.path,
+                                destination: undefined,
+                                isExtract: item.type === "external",
```
jdneo, line 150: The logic here is strange. It's hard to find any relationship between `external` and `isExtract`

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,13 +135,173 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+        }
+        const globs: string[] = await globby(fsPathArray);
+        const sources: IClassPath[] = [];
+        for (const glob of globs) {
+            const tireNode: TrieNode<IUriData> = trie.findFirst(Uri.file(glob).fsPath);
+            if (tireNode === undefined) {
+                continue;
+            }
+            let fsPath = upath.normalizeSafe(Uri.parse(tireNode.value.uri).fsPath);
+            if (!(lstatSync(fsPath).isDirectory())) {
```
jdneo, line 282: async

src/exportJarSteps/GenerateJarExecutor.ts
CsCherrYY: Will rename `isExtract` to `isDependency`

src/exportJarSteps/ExportJarTaskProvider.ts
CsCherrYY: Glob pattern needs a posix path as its input. In windows, simply running `path.normalize()` or `path.posix.normalize()` can't translate the path. Instead of using `upath`, I will write a new utility to translate windows path to posix path.

jdneo: I see. It's ok to use `upath` then.

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
CsCherrYY: When we use the wizard to export jar. We will export a series of folders containing class file. `writeFileRecursively` handle these folders.
When we use a custom task to export jar, we will get the file list (not include any directory) from globby.  We use a trie to save and specific each destination of a particular file, so the `ClassPath` has a valid `destination`. So we simply call `writeFile`.
Do you think it's better to merge the behavior of them and use globby to get the file list when we use the wizard to export jar?

src/exportJarSteps/GenerateJarExecutor.ts
```diff
@@ -133,12 +143,29 @@ export class GenerateJarExecutor implements IExportJarStepExecutor {
...
+                                    destination: undefined,
+                                    isDependency: true,
+                                };
+                                stepMetadata.classpaths.push(classpath);
+                            } else {
+                                const posixPath: string = toPosixPath(item.path);
+                                for (const path of await globby(posixPath)) {
+                                    const classpath: IClassPath = {
+                                        source: path,
+                                        destination: path.substring(posixPath.length + 1),
```
jdneo, line 163: how about path.relative()

jdneo (changes requested):

package.nls.json
```diff
@@ -24,5 +24,8 @@
   "configuration.java.dependency.autoRefresh": "Synchronize Java Projects explorer with changes",
   "configuration.java.dependency.refreshDelay": "The delay time (ms) the auto refresh is invoked when changes are detected",
   "configuration.java.dependency.packagePresentation": "Package presentation mode: flat or hierarchical",
-  "configuration.java.project.exportJar.targetPath": "The default output path of export jar"
+  "configuration.java.project.exportJar.targetPath": "The default output path of export jar",
+  "taskDefinitions.java.project.exportJar.elements": "The content list of export jar.",
+  "taskDefinitions.java.project.exportJar.mainMethod": "The main method in the manifest of exported jar.",
+  "taskDefinitions.java.project.exportJar.targetPath": "The output path of the exported jar. This will override the setting `configuration.java.project.exportJar.targetPath`."
```
jdneo, line 30: Will this get rendered in markdown style? If not, please use " instead of `

src/views/nodeCache/Trie.ts
```diff
@@ -29,14 +29,17 @@ export class Trie<T extends IUriData> {
...
-    public find(fsPath: string): TrieNode<T> | undefined {
+    public find(fsPath: string, returnEarly: boolean = false): TrieNode<T> | undefined {
         let currentNode = this.root;
         const segments: string[] = fsPath.split(path.sep);
 
         for (const segment of segments) {
             if (!segment) {
                 continue;
             }
+            if (returnEarly === true && currentNode.value) {
```
jdneo, line 40: if (returnEarly && currentNode.value ) {
    ...
}

jdneo (changes requested):

src/exportJarSteps/utility.ts
```diff
@@ -81,3 +90,21 @@ export function successMessage(outputFileName: string) {
...
+
+export function toPosixPath(inputPath: string): string {
+    return inputPath.split(win32.sep).join(posix.sep);
+}
+
+export async function getExtensionApi(): Promise<any> {
+    const extension: Extension<any> | undefined = extensions.getExtension("redhat.java");
+    if (extension === undefined) {
+        failMessage("Language Support for Java(TM) by Red Hat isn't running, the export process will be aborted.");
+        return Promise.reject();
```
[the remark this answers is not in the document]
CsCherrYY: Because the caller of this method returns a Promise. The try-catch block which to hand errors is in the caller of this caller. If we use `throw`, it may not be caught by the catch statement.
For this function, I will update a commit, which uses `reject` with error messages and catches them outside instead of using `failMessage` directly.

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,17 +128,162 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+            fsPathArray.push(classPath);
+        }
+        const sources: IClassPath[] = [];
+        for (const glob of await globby(fsPathArray)) {
+            const tireNode: TrieNode<IUriData> = trie.find(Uri.file(platform() === "win32" ? toWinPath(glob) : glob).fsPath, true);
+            if (tireNode === undefined) {
+                continue;
+            }
+            let fsPath = Uri.parse(tireNode.value.uri).fsPath;
+            if (!(await lstat(fsPath)).isDirectory()) {
```
jdneo, line 250: isFile()

jdneo (changes requested):

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,17 +128,162 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+                source: dependency,
+                destination: undefined,
+                isDependency: true,
+            };
+            sources.push(classpath);
+        }
+        return sources;
+    }
+
+    private getClassPathFromMap(matchResult: RegExpMatchArray, source: Map<string, string[]>, projectName: string | undefined): string[] {
```
jdneo, line 273: source -> rawClasspathEntries

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,17 +128,162 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+        const result: string[] = [];
+        if (projectName !== undefined) {
+            for (const classPath of source.get(projectName)) {
+                result.push(this.toAbsolutePosixPath(matchResult.input.replace(matchResult[0], classPath)));
+            }
+        } else {
+            for (const value of source.values()) {
```
jdneo, line 280: value -> classpaths

src/exportJarSteps/GenerateJarExecutor.ts
```diff
@@ -87,22 +91,28 @@ export class GenerateJarExecutor implements IExportJarStepExecutor {
...
             });
         });
         if (_.isEmpty(dependencyItems)) {
             throw new Error("No classpath found. Please make sure your java project is valid.");
         } else if (dependencyItems.length === 1) {
-            stepMetadata.elements.push(dependencyItems[0].path);
+            const classpath: IClassPath = {
+                source: dependencyItems[0].path,
+                destination: undefined,
+                isDependency: false,
```
jdneo, line 113: isDependency -> isArtifact

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,17 +128,162 @@ class ExportJarTaskTerminal implements Pseudoterminal {
     }
 
     public async open(initialDimensions: TerminalDimensions | undefined): Promise<void> {
-        await createJarFile(this.stepMetadata);
-        this.closeEmitter.fire();
+        try {
+            if (!_.isEmpty(this.stepMetadata.elements)) {
+                const runtimeClassPathMap: Map<string, string[]> = new Map<string, string[]>();
```
jdneo, line 133: ClassPath -> Classpath, also please correct all the occurance

src/exportJarSteps/ExportJarTaskProvider.ts
jdneo, line 273: getJarElementsFromClasspathMapping

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -29,14 +45,69 @@ export class ExportJarTaskProvider implements TaskProvider {
...
+                        "${" + ExportJarConstants.TEST_DEPENDENCIES + ":" + project.name + "}");
+                }
+            }
+            const defaultDefinition: IExportJarTaskDefinition = {
+                type: ExportJarTaskProvider.exportJarType,
+                elements: elementList,
+                mainMethod: "",
+                targetPath: ExportJarTargets.DEFAULT_OUTPUT_PATH,
+            };
+            const stepMetadata: IStepMetadata = {
```
jdneo, line 96: move this assignment into

```ts
async (resolvedDefinition: TaskDefinition): Promise<Pseudoterminal> => {
    return new ExportJarTaskTerminal(resolvedDefinition, stepMetadata);
}
```


same in `resolveTask()`

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -29,14 +45,69 @@ export class ExportJarTaskProvider implements TaskProvider {
...
+                        "${" + ExportJarConstants.TESTCOMPILE_OUTPUT + ":" + project.name + "}",
+                        "${" + ExportJarConstants.RUNTIME_DEPENDENCIES + ":" + project.name + "}",
+                        "${" + ExportJarConstants.TEST_DEPENDENCIES + ":" + project.name + "}");
+                }
+            }
+            const defaultDefinition: IExportJarTaskDefinition = {
+                type: ExportJarTaskProvider.exportJarType,
+                elements: elementList,
+                mainMethod: "",
+                targetPath: ExportJarTargets.DEFAULT_OUTPUT_PATH,
```
jdneo, line 94: the default `targetPath` should get from setting

jdneo (changes requested):

package.json
```diff
@@ -476,6 +476,29 @@
         "contents": "You can create a new Java project by clicking the below button.\n[Create new Java project](command:java.project.create)",
         "when": "!java:projectManagerActivated || java:serverMode != Standard"
       }
+    ],
+    "taskDefinitions": [
+      {
+        "type": "java",
+        "required": ["targetPath"],
```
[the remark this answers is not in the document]
CsCherrYY: The original reason is that I can't find a way to use the value of setting `configuration.java.project.exportJar.targetPath` when a user does not set the `targetPath` field. 
I resolve the VS Code variables by passing them into `CustomExecution` as a part of `TaskDefinition`. However, `TaskDefinition` can't be modified in `resolveTask()` in the current VS Code version. See https://github.com/microsoft/vscode/issues/46667#issuecomment-376526692 and https://github.com/microsoft/vscode/issues/58836#issuecomment-500731627
So when we found the `targetPath` field is undifined of the current task, we can't modify it with the setting above and pass it to `CustomExecution` to resolve it.
We have two choices now. One is to set `targetPath` is required, the other is to remain the `targetPath` empty when the user does not set it and open a dialog to ask user the target path when the task is executing. Which one do you prefer?

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -57,17 +128,162 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
+                for (const project of projectList) {
+                    await this.setClasspathMap(project, "runtime", runtimeClassPathMap, runtimeDependencyMap);
+                    await this.setClasspathMap(project, "test", testClassPathMap, testDependencyMap);
+                }
+                this.stepMetadata.classpaths = await this.resolveClassPaths(runtimeClassPathMap,
+                    runtimeDependencyMap, testClassPathMap, testDependencyMap);
+            }
+            await createJarFile(this.stepMetadata);
+        } catch (err) {
+            if (err) {
```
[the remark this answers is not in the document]
CsCherrYY: This is to catch the possible error from `getExtensionApi()` in `setClasspathMap()` of this block. After this PR merged, I will use something like error code to rewrite all the error handing of export jar.

package.json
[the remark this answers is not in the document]
jdneo: let's remove the requirement and ask user if it's not set

src/exportJarSteps/ExportJarTaskProvider.ts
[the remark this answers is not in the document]
jdneo: ok

jdneo:
> "label": "java: exportjar:demo-noMain" // The label of this task to show in the wizard from `Terminal->Run Task...`

Do we have this field?

CsCherrYY:
> > "label": "java: exportjar:demo-noMain" // The label of this task to show in the wizard from `Terminal->Run Task...`
> 
> Do we have this field?

It's a VS Code built-in field.  Its value is in this format: `task.type`: `task.name`

src/exportJarSteps/ExportJarTaskProvider.ts
```diff
@@ -56,18 +126,168 @@ class ExportJarTaskTerminal implements Pseudoterminal {
...
-    elements?: string[];
-    mainMethod?: string;
-    targetPath?: string;
+    private async setClasspathMap(project: INodeData, classpathScope: string,
+                                  outputFolderMap: Map<string, string[]>, artifactMap: Map<string, string[]>): Promise<void> {
+        const extensionApi: any = await getExtensionApi();
+        const classpathResult: IClasspathResult = await extensionApi.getClasspaths(project.uri, { scope: classpathScope });
+        const outputFolders: string[] = [];
+        const artifacts: string[] = [];
+        for (const classpath of classpathResult.classpaths) {
```
jdneo, line 168: for (const classpath of [...classpathResult.classpaths, ...classpathResult.modulepaths])

jdneo (changes requested):
very close, just a few more comments

jdneo:
Some more feedbacks:
image

1. Can we remove `problemMatcher` when generating the task?
2. Default mainMethod should be filled if there is only one main entry
3. `CompileOutput` -> `compileOutput` and same to other reserved words

CsCherrYY:
> Some more feedbacks:
> image
> 
> 1. Can we remove `problemMatcher` when generating the task?
> 2. Default mainMethod should be filled if there is only one main entry
> 3. `CompileOutput` -> `compileOutput` and same to other reserved words

Thanks a lot! Will apply them.

package.nls.json
```diff
@@ -24,5 +24,8 @@
   "configuration.java.dependency.autoRefresh": "Synchronize Java Projects explorer with changes",
   "configuration.java.dependency.refreshDelay": "The delay time (ms) the auto refresh is invoked when changes are detected",
   "configuration.java.dependency.packagePresentation": "Package presentation mode: flat or hierarchical",
-  "configuration.java.project.exportJar.targetPath": "The default output path of export jar"
+  "configuration.java.project.exportJar.targetPath": "The default output path of export jar",
+  "taskDefinitions.java.project.exportJar.elements": "The content list of export jar.",
+  "taskDefinitions.java.project.exportJar.mainMethod": "The main method in the manifest of exported jar.",
+  "taskDefinitions.java.project.exportJar.targetPath": "The output path of the exported jar. This will override configuration.java.project.exportJar.targetPath."
```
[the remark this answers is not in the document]
CsCherrYY: This description is showed when the user hover at the properties like `targetPath` in task.json. I couldn't find a way to add a link to the description here. It seems that only literal text is supported.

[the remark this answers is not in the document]
CsCherrYY: It seems that VS Code does not support markdown description in the `taskDefinitions` currently. So I couldn't find a way to link to the settings in the tasks.json file.

CsCherrYY:
> Some more feedbacks:
> image
> 
> 1. Can we remove `problemMatcher` when generating the task?
> 2. Default mainMethod should be filled if there is only one main entry
> 3. `CompileOutput` -> `compileOutput` and same to other reserved words

It seems that `problemMatcher` cannot be removed programmly. The default `echo` task comes from literal template. Other tasks generated will include a  `problemMatcher` property, whether it is undefined or used.

jdneo:
@CsCherrYY Please resolve the conflict

package.json
```diff
@@ -466,6 +466,28 @@
         "contents": "You can create a new Java project by clicking the below button.\n[Create new Java project](command:java.project.create)",
         "when": "!java:projectManagerActivated || java:serverMode != Standard"
       }
+    ],
+    "taskDefinitions": [
+      {
+        "type": "java",
+        "properties": {
+          "elements": {
```
testforstephen, line 474: A reference about how to support intellisense for constant literals when configuring elements.

https://github.com/microsoft/vscode-java-debug/blob/master/package.json#L346-L373

constant

jdneo: @testforstephen beautiful!

package.nls.json
```diff
@@ -24,5 +24,13 @@
   "configuration.java.dependency.autoRefresh": "Synchronize Java Projects explorer with changes",
   "configuration.java.dependency.refreshDelay": "The delay time (ms) the auto refresh is invoked when changes are detected",
   "configuration.java.dependency.packagePresentation": "Package presentation mode: flat or hierarchical",
-  "configuration.java.project.exportJar.targetPath": "The default output path of export jar"
+  "configuration.java.project.exportJar.targetPath": "The default output path of export jar.",
+  "taskDefinitions.java.project.exportJar.label": "The label of export jar task.",
+  "taskDefinitions.java.project.exportJar.elements": "The content list of the exported jar.",
+  "taskDefinitions.java.project.exportJar.mainClass": "The main class in the manifest of the exported jar.",
+  "taskDefinitions.java.project.exportJar.targetPath": "The output path of the exported jar. This will override configuration.java.project.exportJar.targetPath.",
+  "taskDefinitions.java.project.exportJar.compileOutput": "All of the output folders in the runtime scope.",
```
CsCherrYY, line 32: @jdneo What do you think of the wording here, including `output folders` and `dependencies` ? Are them the proper ones? Or could you please give some suggestions?

jdneo (approved):
Thanks @CsCherrYY

PR #353, opened 2020-10-16
feat: Add 'rename' into the view context menu

xqzlgy2:
Enable User to rename package/file in context menu like below.

- User input new file name through input box
- If file extension not provided, use previous extension name by default, otherwise use user's input
- Refactoring will triggered automatically by language server
- Register key binding for rename
- Also add delete/rename key bindings for mac

part of #207 

image
image

### DIFF
--- a/src/explorerCommands/rename.ts
+++ b/src/explorerCommands/rename.ts
@@ -15,7 +15,7 @@ export async function renameFile(node: DataNode, selectedNode: ExplorerNode): Pr
             return;
         }
     }
-    
+
     const newName: string | undefined = await window.showInputBox({
         placeHolder: "Input new file name",
         ignoreFocusOut: true,
--- a/src/explorerCommands/rename.ts
+++ b/src/explorerCommands/rename.ts
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as path from "path";
+import { Uri, window, workspace, WorkspaceEdit } from "vscode";
+import { DataNode } from "../views/dataNode";
+
+export async function renameFile(node: DataNode): Promise<void> {
+    const newName: string | undefined = await window.showInputBox({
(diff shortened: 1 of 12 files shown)


src/explorerCommands/rename.ts
```diff
@@ -0,0 +1,43 @@
...
+    const renamedFilePath = getRenamedFilePath(oldFsPath, newName);
+
+    const workspaceEdit: WorkspaceEdit = new WorkspaceEdit();
+    workspaceEdit.renameFile(Uri.file(oldFsPath), Uri.file(renamedFilePath));
+    workspace.applyEdit(workspaceEdit);
+}
+
+function getRenamedFilePath(oldUri: string, newName: string): string {
+    // preserve default file extension if not provided
+    if (!path.extname(newName)) {
```
jdneo, line 38: what will happen if the rename happens on a file which does not have a file extension name originally? For example, rename file `A` -> `B`

jdneo:
@0dinD, just FYI this is the PR for `rename`

src/explorerCommands/rename.ts
xqzlgy2: It will behave normally, since `path.extname` will return an empty string if no extension name provided.

image

src/explorerCommands/rename.ts
```diff
@@ -0,0 +1,43 @@
...
+export async function renameFile(node: DataNode, selectedNode: ExplorerNode): Promise<void> {
+    // if command not invoked by context menu, use selected node in explorer
+    if (!node) {
+        node = selectedNode as DataNode;
+        if (!isMutable(node)) {
+            return;
+        }
+    }
+
+    const newName: string | undefined = await window.showInputBox({
```
jdneo, line 19: Pre-fill the current file name in the input box.

jdneo (approved):

jdneo:
Thanks @xqzlgy2!

PR #357, opened 2020-10-19
Check if the selected item is valid when quickpick

### LINKED ISSUE #358: Prevent exception throwing when invalid value is inputted during export jar
For example, current code block in `ResolveMainMethodExecutor.ts`:
```
pickBox.onDidAccept(() => {
    stepMetadata.mainMethod = pickBox.selectedItems[0].description;
    stepMetadata.steps.push(ExportJarStep.ResolveMainMethod);
    return resolve(true);
}
```
if the user input some invalid values, making `pickBox.selectedItems` is empty. Visiting `pickBox.selectedItems[0].description` will result in an exception.

CsCherrYY:
fix #358

### DIFF
--- a/src/exportJarSteps/GenerateJarExecutor.ts
+++ b/src/exportJarSteps/GenerateJarExecutor.ts
@@ -121,35 +121,34 @@ export class GenerateJarExecutor implements IExportJarStepExecutor {
             }
         }
         const disposables: Disposable[] = [];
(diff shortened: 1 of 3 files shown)


jdneo:
If this is to fix a bug, please also file an issue for it.

jdneo (approved):

PR #360, opened 2020-10-20
Upgrade the icon of export jar

CsCherrYY:


jdneo (changes requested):
Please also update the dependency for the vscode type definition to 1.50.0 and update the package.lock.json

jdneo (approved):

jdneo:
Thanks @CsCherrYY!

PR #361, opened 2020-10-20
test: Add test cases for new context value

xqzlgy2:
Since we have added `File/Type` in `ContextValueType`, it's necessary to add test cases for them.
Tested context value in this PR:
- `java:file+uri`
- `java:type+class+uri`
- `java:type+enum+uri`
- `java:type+interface+uri`


jdneo (approved):

jdneo:
Thanks @xqzlgy2!

PR #365, opened 2020-10-30
feat: Add context value for folder

xqzlgy2:
- Add new type `Folder` into `ContextValueType`
- Also enabled context menu for non-package folder and resource packageRoot
- Test case for new context value


jdneo (approved):

jdneo:
Thanks @xqzlgy2!

PR #366, opened 2020-11-03
fix: Validate input & change pre-fill value

xqzlgy2:
This PR fixed two parts:

1. Input validation:
- If user rename `PrimaryTypeNode`, then:
    1) Input should be a legal Java identifier 
- If user rename `packageRoot` or `packageRootNode`, then:
    1) Input should also be a legal identifier
    3) Could not contain `.`, due to limitations, currently we can only rename innermost package

2. Pre-fill value:
- For package, since cross-level rename is not supported, we only show innermost folder name
- For `PrimaryTypeNode`, only show class name
- For other files, show extension name, but not pre-selected

### DIFF
--- a/src/explorerCommands/new.ts
+++ b/src/explorerCommands/new.ts
@@ -138,7 +138,7 @@ function getNewPackagePath(packageRootPath: string, packageName: string): string
     return path.join(packageRootPath, ...packageName.split("."));
 }
 
-function checkJavaQualifiedName(value: string): string {
+export function checkJavaQualifiedName(value: string): string {
     if (!value || !value.trim()) {
         return "Input cannot be empty.";
     }
--- a/src/explorerCommands/new.ts
+++ b/src/explorerCommands/new.ts
@@ -5,8 +5,8 @@ import * as fse from "fs-extra";
 import * as path from "path";
 import { QuickPickItem, Uri, window, workspace, WorkspaceEdit } from "vscode";
 import { NodeKind } from "../java/nodeData";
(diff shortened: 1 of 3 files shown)


src/explorerCommands/rename.ts
```diff
@@ -16,29 +19,85 @@ export async function renameFile(node: DataNode, selectedNode: ExplorerNode): Pr
...
         ignoreFocusOut: true,
+        valueSelection: getValueSelection(node.uri),
+        validateInput: async (value: string): Promise<string> => {
+            const checkMessage = CheckQualifiedInputName(value, node.nodeData.kind);
+            if (checkMessage) {
+                return checkMessage;
+            }
+
+            if (await fse.pathExists(getRenamedFsPath(oldFsPath, value))) {
+                return "Class/Package already exists.";
```
jdneo, line 36: File path: 'xxxx' already exists.

src/explorerCommands/new.ts
```diff
@@ -138,7 +138,7 @@ function getNewPackagePath(packageRootPath: string, packageName: string): string
     return path.join(packageRootPath, ...packageName.split("."));
 }
 
-function checkJavaQualifiedName(value: string): string {
+export function checkJavaQualifiedName(value: string): string {
```
jdneo, line 141: move to util

jdneo (changes requested):

jdneo (approved):

jdneo:
Thanks @xqzlgy2

PR #375, opened 2020-11-16
chore: Send event for 'query-expfeature'

jdneo:


Eskibear (approved):

jdneo:
@daytonellwanger Will be appreciated if you can take a look.

src/ExperimentationService.ts
```diff
@@ -13,7 +12,12 @@ class ExperimentationTelemetry implements IExperimentationTelemetry {
     }
 
     public postEvent(eventName: string, props: Map<string, string>): void {
-        // do nothing
+        const payload: any = { __event_name__: eventName };
+        for (const [key, value] of props) {
```
daytonellwanger, line 16: Does `payload = { ...props, ___event_name___: eventName }` work with the `Map<string, string>` type?

jdneo: Seems doesn't, but in `ES2019` we can write it as `payload = { ...Object.fromEntries (props), ___event_name___: eventName }`

PR #376, opened 2020-11-16
fix: Cannot show the content of a referenced jar

### LINKED ISSUE #370: Can't show the class in the reference jar
### Environment:

- Vscode: 1.50.1
- JDK: 11.0.8
- OS: Win/Mac/Linux
- Plugin version: 0.14.0

### Repro steps:

1. Open invisible project: https://github.com/jdneo/invisble-project
2. Add jars in the test-lib folder
3. Verify class of jar which under `Referenced Libraries`

### Result:
Can't show the class in the reference jar
image

### Expect:
image

jdneo:
fix #370

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
@@ -341,7 +341,7 @@ public class PackageCommand {
 
     private static List<PackageNode> getPackages(PackageParams query, IProgressMonitor pm) {
         try {
-            IPackageFragmentRoot packageRoot = (IPackageFragmentRoot) JavaCore.create(query.getHandlerIdentifier());
+            IPackageFragmentRoot packageRoot = getPackageFragmentRootFromQuery(query);
             if (packageRoot == null) {
                 throw new CoreException(
                         new Status(IStatus.ERROR, JdtlsExtActivator.PLUGIN_ID, String.format("No package root found for %s", query.getPath())));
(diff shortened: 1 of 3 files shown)


jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
```diff
@@ -354,6 +354,25 @@
         return Collections.emptyList();
     }
 
+    private static IPackageFragmentRoot getPackageFragmentRootFromQuery(PackageParams query) {
+        IPackageFragmentRoot packageRoot = (IPackageFragmentRoot) JavaCore.create(query.getHandlerIdentifier());
+        if (packageRoot != null) {
+            return packageRoot;
+        }
+
+        // jar in Referenced Libraries must be constructed from path
```
Eskibear, line 363: When will ` JavaCore.create(query.getHandlerIdentifier())` return null?

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
```diff
@@ -354,6 +354,25 @@
...
+        IJavaProject javaProject = getJavaProject(query.getProjectUri());
+        if (javaProject != null) {
+            try {
+                packageRoot = javaProject.findPackageFragmentRoot(Path.fromPortableString(query.getRootPath()));
```
CsCherrYY, line 367: What is the difference between `fromOSString()` and `fromPortableString()` here?

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
jdneo: e.g. when the input is null

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
jdneo: `fromPortableString()` is used here because the path is generated by `toPortableString()`

See: https://github.com/microsoft/vscode-java-dependency/blob/master/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java#L252

CsCherrYY (approved):

PR #384, opened 2020-11-17
Fix: Target path is wrong when exporting jar from a workspace node

### LINKED ISSUE #382: Target path is wrong when exporting jar from a workspace node
Currently, if the user tries to export jar button from a workspace node, as shown in the following screenshot:

Screenshot 2020-11-17 143809

If the setting `java.project.exportJar.targetPath` is set to `${workspaceFolder}/${workspaceFolderBasename}.jar`, the output folder will always be the first workspace folder, regardless of the entry node.

CsCherrYY:
Fix #382


src/exportJarFileCommand.ts
```diff
@@ -36,17 +36,30 @@ export async function executeExportJarTask(node?: INodeData): Promise<void> {
...
+    try {
+        await stepMap.get(ExportJarStep.ResolveJavaProject).execute(stepMetadata);
+    } catch (err) {
+        if (err) {
+            failMessage(`${err}`);
+        }
+        return;
+    }
     tasks.executeTask(ExportJarTaskProvider.getTask(stepMetadata));
     return;
```
jdneo, line 48: This `return` is redundent

jdneo (changes requested):

src/exportJarFileCommand.ts
```diff
@@ -32,33 +32,26 @@ export async function executeExportJarTask(node?: INodeData): Promise<void> {
...
+        isExportingJar = false;
+        return;
+    }
+    tasks.executeTask(ExportJarTaskProvider.getTask(stepMetadata));
+}
+
+export async function finishExportJarTask(restart: boolean, node?: INodeData): Promise<void> {
+    isExportingJar = false;
+    if (restart) {
+        executeExportJarTask(node);
```
CsCherrYY, line 55: Currently, the `node` here is always undefined because only the export jar process triggered by the button in the title of the Project Manager (not by the inline button of workspaceNode) could go back to the `ResolveJavaProject` step.
Should the `node` be `undefined` or just keep it for the consistency here?

jdneo: This function is a trouble. Some reasons:

- Still, it makes the people have to read the code between `exportJarFileCommand.ts` & `ExportJarTaskProvider.ts` back and forth
- It's named `finishExportJarTask` but with a param `restart`, confusing here.

jdneo (approved):

PR #386, opened 2020-11-20
test: Add test cases for different projects

xqzlgy2:
This PR added tests for Simple/Maven/Gradle projects.
For each kind of project, two cases are tested:
- Can node render correctly: from root node, test nodes' name and number layer by layer
- Can node have correct uri: for each node, convert it's uri to fspath, and compare with correct ones

Also updated and added dependencies.


test/gradle-suite/index.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as glob from "glob";
+import * as Mocha from "mocha";
+import * as path from "path";
+
+export function run(): Promise<void> {
+    // Create the mocha test
+    const mocha = new Mocha({
+        ui: "tdd",
+    });
+    mocha.useColors(true);
```
jdneo, line 13: Can it be set in constructor?

test/gradle-suite/projectView.test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import * as assert from "assert";
+import { commands, extensions } from "vscode";
+import { Commands, ContainerNode, contextManager, DependencyExplorer,
+    PackageRootNode, PrimaryTypeNode, ProjectNode } from "../../extension.bundle";
+import { fsPath, Uris } from "../shared";
+
+suite("Gradle Project View Tests", () => {
+    test("Can node render correctly", async function() {
+        this.timeout(1 * 60 * 1000);
```
jdneo, line 12: Why not set timeout as a parameter for mocha instance

test/gradle-suite/projectView.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        const mainClass = mainClasses[0] as PrimaryTypeNode;
+        assert.equal(mainClass.name, "GradleTest", "Name of main class should be \"GradleTest\"");
+    });
+
+    test("Can node have correct uri", async function() {
+        this.timeout(1 * 60 * 1000);
+        await extensions.getExtension("vscjava.vscode-java-dependency")!.activate();
+
+        // context would be initialized after this command
+        await commands.executeCommand(Commands.JAVA_PROJECT_ACTIVATE);
```
jdneo, line 48: Same logic can be moved into `suiteSetup`

jdneo (changes requested):

.vscodeignore
```diff
@@ -9,11 +9,12 @@ vscode/**
 tsconfig.json
 gulpfile.js
 .gitignore
-dist/**/*.map
+dist/**/test
 webpack.*.js
 tslint.json
 node_modules
 scripts
+extension.bundle.ts
```
jdneo, line 17: Let's also add `javaConfig.json` to the ignore list

jdneo (approved):
LGTM

PR #388, opened 2020-11-23
chore: Fix error in CI test process

xqzlgy2:
This PR fix error in current CI test process.
Test will not begin until language server is switched to the Standard mode.
Also fixed previous problem in CI (OSX will hang in build process).

### DIFF
--- a/test/shared.ts
+++ b/test/shared.ts
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import * as assert from "assert";
 import * as path from "path";
-import { Uri } from "vscode";
-import { DataNode } from "../extension.bundle";
+import { commands, Event, extensions, Uri } from "vscode";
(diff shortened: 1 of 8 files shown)


test/shared.ts
```diff
@@ -37,3 +37,21 @@ export function truePath(...paths: string[]) {
     const basePath = path.join(__dirname, "..", "..", "test");
     return path.join(basePath, ...paths);
 }
+
+export async function setupTestEnv() {
+    const extension = await extensions.getExtension("redhat.java")!.activate();
```
jdneo, line 42: No need to activate here

jdneo (changes requested):

test/shared.ts
xqzlgy2: But the promise below will keep waiting until LS is ready ins standard mode. If we do not activate, are there chances that whole process will stuck?

jdneo (approved):
Thanks @xqzlgy2!

PR #391, opened 2020-11-26
feat: Update project from project node in explorer

jdneo:


src/views/dependencyDataProvider.ts
```diff
@@ -55,6 +55,17 @@ export class DependencyDataProvider implements TreeDataProvider<ExplorerNode> {
             commands.executeCommand(Commands.JAVA_BUILD_WORKSPACE)));
         context.subscriptions.push(instrumentOperationAsVsCodeCommand(Commands.JAVA_PROJECT_CLEAN_WORKSPACE, () =>
             commands.executeCommand(Commands.JAVA_CLEAN_WORKSPACE)));
+        context.subscriptions.push(instrumentOperationAsVsCodeCommand(Commands.JAVA_PROJECT_UPDATE, async (node: INodeData) => {
+            if (!node.uri) {
+                return;
```
Eskibear, line 60: 1. In what case, node.uri is undefined?
2. In such case, when user click "Update", it simply returns silently? Is it a good experience?

jdneo: 1. Based on the current code logic, this won't happen. 
2. Any suggestions?

Eskibear: 1. ok, then it's fine. 
2. I would suggest to throw an error/warning here, then users would know the action didn't take effect.  (if for some reason node.uri is wrongly set to undefined)

jdneo: I see. Will show a error message instead

Eskibear (approved):

PR #393, opened 2020-11-27
feat: Add more keybindings for context menu

xqzlgy2:
Add keybindings for `Reveal in Explorer`, `Copy Path`, `Copy Relative Path`.

image
image


jdneo (approved):
Thanks @xqzlgy2

PR #401, opened 2020-12-07
fix: Prevent multiple calls of reveal() at the same time

### LINKED ISSUE #400: Duplicated items show in the explorer
image

This is commonly observed when multiple calls for the `reveal()` happens then the view will be rendered repeatedly. We need to find a way to eliminate the duplicated calls

jdneo:
fix #400

### DIFF
--- a/src/views/dependencyExplorer.ts
+++ b/src/views/dependencyExplorer.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 import * as fse from "fs-extra";
+import * as _ from "lodash";
 import * as path from "path";
 import { commands, Disposable, ExtensionContext, TextEditor, TreeView, TreeViewVisibilityChangeEvent, Uri, window } from "vscode";
 import { instrumentOperationAsVsCodeCommand } from "vscode-extension-telemetry-wrapper";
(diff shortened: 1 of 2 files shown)


src/views/dependencyExplorer.ts
```diff
@@ -137,17 +140,28 @@ export class DependencyExplorer implements Disposable {
...
             const paths: INodeData[] = await Jdtls.resolvePath(uri.toString());
-            if (!paths || paths.length === 0) {
-                return;
+            if (!_.isEmpty(paths)) {
+                node = await this._dataProvider.revealPaths(paths);
             }
-            node = await this._dataProvider.revealPaths(paths);
         }
 
-        if (this._dependencyViewer.visible) {
```
CsCherrYY, line 149: Is `this._dependencyViewer.visible` will be always `ture`? Or it should not be a condition for `reveal`?

jdneo: According to the document for the `reveal()` API:

> Reveals the given element in the tree view. If the tree view is not visible then the tree view is shown and element is revealed.

CsCherrYY (approved):

PR #417, opened 2020-12-14
fix: NPE error when open non-project file

### LINKED ISSUE #418: NPE error when open non-project file
When we open a non-java file which doesn't belong to the opened project (e.g. execute `Java: Open Java Language Server Log`), LS will throw NPE error like below:

image
image

xqzlgy2:
When we open a non-java file which doesn't belong to the opened project, LS will throw NPE error, since in `resolvePath` api we assume the file should be in workspace. So, additional checks are needed.

fix #418

### DIFF
--- a/src/views/dependencyExplorer.ts
+++ b/src/views/dependencyExplorer.ts
@@ -4,7 +4,7 @@
 import * as fse from "fs-extra";
 import * as _ from "lodash";
 import * as path from "path";
-import { commands, Disposable, ExtensionContext, TextEditor, TreeView, TreeViewVisibilityChangeEvent, Uri, window, workspace } from "vscode";
+import { commands, Disposable, ExtensionContext, TextEditor, TreeView, TreeViewVisibilityChangeEvent, Uri, window } from "vscode";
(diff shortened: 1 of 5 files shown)


jdneo:
I think you can check this case at the client side in order to save the traffic to the language server: https://github.com/microsoft/vscode-java-dependency/blob/master/src/views/dependencyExplorer.ts#L49

Also please create a new issue for this bug for easy tracking.

Thanks.

src/views/dependencyExplorer.ts
```diff
@@ -48,6 +48,10 @@ export class DependencyExplorer implements Disposable {
             window.onDidChangeActiveTextEditor((textEditor: TextEditor) => {
                 if (this._dependencyViewer.visible && textEditor && textEditor.document && Settings.syncWithFolderExplorer()) {
                     const uri: Uri = textEditor.document.uri;
+                    // if editor doesn't belong to workspace, do nothing
+                    if (uri.fsPath === workspace.asRelativePath(uri.fsPath)) {
```
jdneo, line 52: Using `workspace.getWorkspaceFolder()` is more readable IMO

jdneo (approved):

PR #426, opened 2020-12-18
feat: Make openFile experience consistent with built-in trees

xqzlgy2:
This PR changes two parts:

1. From VS Code 1.52, all custom tree behavior has been changed, which enable our plugin to have more consistent experience with file explorer when open file. That means, all following interactions are enabled for `fileNode` and `primaryTypeNode`:

- Double-click or mouse-middle-click to open in non-preview.
- Enter to open in non-preview mode (Space to open in preview mode).
- Alt+Click to open an editor to the side of the active one.

reference: [https://code.visualstudio.com/updates/v1_52](ref)

2. Since changes above needs to register `vscode.open` as TreeItem's command id, we can not use customized `java.view.package.openFile` command to send telemetry. As compensate, we send these additional info in telemetry:

- `{projectManagerVisible: 1}` when project manager is expanded
- counter for `didChangeSelection`, `didCollapseElement`, `didExpandElement` events

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -20,8 +20,6 @@ export namespace Commands {
 
     export const VIEW_PACKAGE_REFRESH = "java.view.package.refresh";
 
-    export const VIEW_PACKAGE_OPEN_FILE = "java.view.package.openFile";
-
     export const VIEW_PACKAGE_OUTLINE = "java.view.package.outline";
 
     export const VIEW_PACKAGE_REVEAL_FILE_OS = "java.view.package.revealFileInOS";
--- a/src/utility.ts
+++ b/src/utility.ts
@@ -5,6 +5,7 @@ import { Uri, window, workspace, WorkspaceFolder } from "vscode";
 import { setUserError } from "vscode-extension-telemetry-wrapper";
 import { languageServerApiManager } from "./languageServerApi/languageServerApiManager";
 import { Settings } from "./settings";
+import { Lock } from "./utils/Lock";
 
 export class Utility {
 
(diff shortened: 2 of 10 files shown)


src/commands.ts
```diff
@@ -20,8 +20,6 @@ export namespace Commands {
 
     export const VIEW_PACKAGE_REFRESH = "java.view.package.refresh";
 
-    export const VIEW_PACKAGE_OPEN_FILE = "java.view.package.openFile";
-
     export const VIEW_PACKAGE_OUTLINE = "java.view.package.outline";
```
jdneo, line 23: Just curious, we have so many constants in this file, is there anyone that is not referenced but still exists in this file?

src/utility.ts
```diff
@@ -38,6 +39,22 @@ export class Utility {
 
 }
 
+export class EventCounter {
+    public static dict: {[key: string]: number} = {};
+
+    public static async increase(event: string) {
+        try {
+            await this._lock.acquire();
```
jdneo, line 47: why do we need the lock here?

src/views/dependencyExplorer.ts
```diff
@@ -77,13 +82,32 @@ export class DependencyExplorer implements Disposable {
...
+                    await commands.executeCommand(Commands.VSCODE_OPEN, uri, { preserveFocus: true });
                 }
 
                 this.reveal(uri);
             }),
         );
 
+        // register telemetry events
+        context.subscriptions.push(
+            this._dependencyViewer.onDidChangeSelection(async (_e: TreeViewSelectionChangeEvent<ExplorerNode>) => {
```
jdneo, line 94: you can put all the three listeners one invocation of `subscription.push()`

jdneo (changes requested):
I think we need also update the vscode engine version to 1.52.0 to limit the impact.

src/commands.ts
xqzlgy2: Just did a check, all command constants are used at least once.

src/utility.ts
xqzlgy2: Now it works fine without lock, I add it just in case of asynchronous operations in future. For example, if we want to reuse this class, and add `decrease` method to this class, a lock would make it safer.

jdneo: If there is no async invocations, there is no need to do that, given that js is single thread. You can check about the `execution stack, event loop` in js for more details.

jdneo (approved):

jdneo:
Thanks @xqzlgy2

PR #434, opened 2021-01-05
Fix add libraries dialog not working on Linux

### LINKED ISSUE #238: Cannot select any directory when adding Referenced Libraries, and cannot add it again after deleting
**Configuration:**
Vscode: 1.42.1
JDK:1.8.0_241
OS: Linux
Build: 0.8.0

**Repro steps:**
1.Open Folder
2.Add a jar file or a folder to project classpath
3. Delete the added content
4. Repeat step 2

Result:
1.The user cannot select the desired folder and the directory cannot go forward or backward.
1be86224-1d21-4e39-9023-f7b55e923b6a

2.After deleting the added a jar file or a folder, it cannot be added again
image

0dinD:
Fixes #238

When trying to add a library to `Referenced Libraries` for invisible projects on Linux, all files are grayed out (unselectable), and you can't even navigate into folders. This makes it impossible to add libraries for invisible projects on Linux, unless you modify the settings.json manually.

This is caused by the unclear API docs for [OpenDialogOptions](https://code.visualstudio.com/api/references/vscode-api#OpenDialogOptions), which say:
> A dialog can select files, folders, or both. This is not true for Windows which enforces to open either files or folder, but *not both*.

This extension then assumes that `canSelectFiles` and `canSelectFolders` can be activated at the same time on both macOS and Linux - but it's actually unsupported on both Windows *and Linux* (which is the root cause of the issue).

Behind the scenes, VS Code is using the Electron API to show the file dialog, and the whole situation actually becomes much clearer if you read the [Electron docs for `dialog`](https://www.electronjs.org/docs/api/dialog#dialogshowopendialogbrowserwindow-options):
> **Note:** On Windows and Linux an open dialog can not be both a file selector and a directory selector, so if you set `properties` to `['openFile', 'openDirectory']` on these platforms, a directory selector will be shown.

So what currently happens is that this extension requests a file dialog that can select both files and folders on Linux, which Electron then converts to a file dialog that can **only** select folders. That means you can't select any Jar files, and you also can't select (or navigate into) any folders, because they don't match the `jar` file filter.

30ab944 is a very simple fix that just makes sure to only offer directory selection on macOS.

But after implementing this, I felt like it was a bit of a shame that Windows and Linux users were missing this QOL feature, just because of some Electron or OS limitation. So in b2d3702 I added a separate button for adding folders to the `Referenced Libraries`, so that users can also add library folders on Windows and Linux.

Let me know if these changes sound good, and if the naming makes sense for all the buttons/tooltips. I would also like some help with the [Chinese translation for the new/updated button tooltips](https://github.com/0dinD/vscode-java-dependency/commit/b2d3702a40079302578a744877b0fcd68beabce6#diff-1987abce949e250b7899192a0953c3f408d71048617f24c4a91c4d27ce4c69f5), since I don't know any (and I don't really trust Google Translate to do a good job).

### DIFF
--- a/src/controllers/libraryController.ts
+++ b/src/controllers/libraryController.ts
@@ -4,6 +4,7 @@
 import * as fse from "fs-extra";
 import * as _ from "lodash";
 import * as minimatch from "minimatch";
+import { platform } from "os";
 import * as path from "path";
 import { Disposable, ExtensionContext, Uri, window, workspace, WorkspaceFolder } from "vscode";
 import { instrumentOperationAsVsCodeCommand } from "vscode-extension-telemetry-wrapper";
@@ -35,13 +36,13 @@ export class LibraryController implements Disposable {
         if (!libraryGlobs) {
             libraryGlobs = [];
             const workspaceFolder: WorkspaceFolder | undefined = Utility.getDefaultWorkspaceFolder();
-            const isWindows = process.platform.indexOf("win") === 0;
+            const isMac = platform() === "darwin";
             const results: Uri[] | undefined = await window.showOpenDialog({
                 defaultUri: workspaceFolder && workspaceFolder.uri,
                 canSelectFiles: true,
-                canSelectFolders: isWindows ? false : true,
+                canSelectFolders: isMac ? true : false,
                 canSelectMany: true,
-                openLabel: isWindows ? "Select jar files" : "Select jar files or directories",
+                openLabel: isMac ? "Select jar files or directories" : "Select jar files",
                 filters: { Library: ["jar"] },
             });
             if (!results) {
--- a/src/controllers/libraryController.ts
+++ b/src/controllers/libraryController.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import * as fse from "fs-extra";
 import * as _ from "lodash";
 import * as minimatch from "minimatch";
 import { platform } from "os";
@@ -57,7 +58,8 @@ export class LibraryController implements Disposable {
             // keep the param: `includeWorkspaceFolder` to false here
             // since the multi-root is not supported well for invisible projects
             const uriPath = workspace.asRelativePath(uri, false);
-            return canSelectFolders ? uriPath + "/**/*.jar" : uriPath;
+            const isLibraryFolder = canSelectFolders || isMac && (await fse.stat(uri.fsPath)).isDirectory();
+            return isLibraryFolder ? uriPath + "/**/*.jar" : uriPath;
         })));
     }
 
--- a/src/controllers/libraryController.ts
+++ b/src/controllers/libraryController.ts
@@ -1,10 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-import * as fse from "fs-extra";
 import * as _ from "lodash";
 import * as minimatch from "minimatch";
-import { platform } from "os";
 import * as path from "path";
 import { Disposable, ExtensionContext, Uri, window, workspace, WorkspaceFolder } from "vscode";
 import { instrumentOperationAsVsCodeCommand } from "vscode-extension-telemetry-wrapper";
@@ -21,6 +19,7 @@ export class LibraryController implements Disposable {
     public constructor(public readonly context: ExtensionContext) {
         this.disposable = Disposable.from(
             instrumentOperationAsVsCodeCommand(Commands.JAVA_PROJECT_ADD_LIBRARIES, () => this.addLibraries()),
+            instrumentOperationAsVsCodeCommand(Commands.JAVA_PROJECT_ADD_LIBRARY_FOLDERS, () => this.addLibraries(true)),
             instrumentOperationAsVsCodeCommand(Commands.JAVA_PROJECT_REMOVE_LIBRARY, (node: DataNode) =>
                 node.uri && this.removeLibrary(Uri.parse(node.uri).fsPath)),
             instrumentOperationAsVsCodeCommand(Commands.JAVA_PROJECT_REFRESH_LIBRARIES, () =>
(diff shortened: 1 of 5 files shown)


0dinD:
Not sure why the test failed on Linux, the errors in the log don't seem to be related to the changes in this PR. And I've ran the tests successfully on my development machine (running Debian 10).

jdneo:
Thank you @0dinD, I'll take a look!

jdneo:
@0dinD Thank you for the detailed explanation. And the fix looks promising!

To me, personally I prefer the way 30ab944 does for this issue. b2d3702 is also great, but it introduces a new command and buttons . I'm not sure how users can easily get the point between `add file` and `add folder`.

So I tend to keep it simple:

- mac: files and folders are both fine
- linux & win: files only

But, we can definitely revisit this if we see more and more users are asking adding folder support on Linux and Windows. What do you think?

0dinD:
@jdneo I do agree that keeping it simple is preferable over adding a new button.

An alternative solution that I just came to think of is to make the `java.project.addLibraryFolders` command the "alt" action for the same button ([see the docs](https://code.visualstudio.com/api/references/contribution-points#contributes.menus)). This way, most users will only see the normal `addLibraries` button, while more advanced users who know about the alt command can still easily add "library folders". What do you think about that solution?

jdneo:
@0dinD I think that's a good idea. Will you update the PR for that?

0dinD:
@jdneo I'll update the PR later today. But since it still requires two separate commands, I do still need translations for the new command titles. Are you able to help with that?

jdneo:
No worry, I'll append my commit to this PR :)

0dinD:
OK, I've addressed the changes we discussed in 4665d8c.

src/controllers/libraryController.ts
```diff
@@ -31,36 +31,34 @@ export class LibraryController implements Disposable {
...
-        }
-
+    public async addLibraryGlobs(libraryGlobs: string[]) {
         const setting = Settings.referencedLibraries();
         setting.exclude = this.dedupAlreadyCoveredPattern(libraryGlobs, ...setting.exclude);
         setting.include = this.updatePatternArray(setting.include, ...libraryGlobs);
         Settings.updateReferencedLibraries(setting);
     }
 
+    public async addLibraries(folders?: boolean) {
```
jdneo, line 41: Change the param name to `canSelectFolders` to just align with the API?

src/controllers/libraryController.ts
```diff
@@ -31,36 +31,34 @@ export class LibraryController implements Disposable {
...
+        const workspaceFolder: WorkspaceFolder | undefined = Utility.getDefaultWorkspaceFolder();
+        const results: Uri[] | undefined = await window.showOpenDialog({
+            defaultUri: workspaceFolder && workspaceFolder.uri,
+            canSelectFiles: !folders,
+            canSelectFolders: folders,
```
jdneo, line 46: One possible regression here is: for the Mac users, they actually lose the capability to select both files and folders in one dialog.

So maybe still also check if it is in MacOS?

src/controllers/libraryController.ts
0dinD: Makes sense, fixed in a735dc5.

src/controllers/libraryController.ts
0dinD: I actually think it's better for the sake of simplicity to keep folder selection specific to the more advanced alt command. Since I don't think we can change the command title based on OS (without creating a third command that only displays on macOS), there is no consistent way to communicate with the user that *both* Jar files and folders can be selected when running macOS. However, I do agree that removing the capability is a regression (which might bother some users), so I added back folder selection to the `addLibraries` dialog on macOS in a735dc5.

src/controllers/libraryController.ts
```diff
@@ -31,36 +32,35 @@ export class LibraryController implements Disposable {
...
+            filters: canSelectFolders ? { Folders: ["*"] } : { "Jar Files": ["jar"] },
+        });
+        if (!results) {
+            return;
+        }
+        this.addLibraryGlobs(await Promise.all(results.map(async (uri: Uri) => {
+            // keep the param: `includeWorkspaceFolder` to false here
+            // since the multi-root is not supported well for invisible projects
+            const uriPath = workspace.asRelativePath(uri, false);
+            return canSelectFolders ? uriPath + "/**/*.jar" : uriPath;
```
jdneo, line 60: When it's on MacOS, and triggered from `addLibraries` (`canSelectFolders` is `undefined`), if the user selects a folder, then it will not append `/**/*.jar`.

SO maybe we still need the previous logic -- check if it is a folder or not and return different value accordingly.

0dinD: Good catch! I missed that since I don't have a mac to test on, but I'll add the old logic back.

jdneo:
Chinese translation added.

BTW, I update `Remove Library from Project Classpath` to `Remove from Project Classpath` since I think there is no need to say remove 'what' in the command title.

src/controllers/libraryController.ts
0dinD: Fixed in 2197367 (although I can't test if it works, since I don't have a mac).

jdneo (approved):
Verified on Mac and Windows, all works fine!

Thank you @0dinD !

0dinD:
No problem, glad I could help!

FYI @jdneo the issue fixed in this PR seems to be related to a recent negative review on the [Java Extension Pack](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-pack&ssr=false#review-details) and the [Language Support for Java](https://marketplace.visualstudio.com/items?itemName=redhat.java&ssr=false#review-details) extension. I can see you've already referenced this PR on the [review left on this extension](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-dependency&ssr=false#review-details), but it would probably be a good idea to respond to all reviews (even though they're from the same user), as it might help others. I don't know about the extension pack, but I'm guessing @fbricon or @rgrunber is responsible for the Java extension.

PR #437, opened 2021-01-07
feat: Add test metadata to the nodes

jdneo:
This PR add `+test` attributes to the test nodes.

If a package fragment root has following attributes, then it is marked as a test root:
- `test=true`
- `maven.scope` includes `test`
- `gradle_scope` includes `test`

When a package fragment root is a test root, then itself and all of its children will have a test attribute `+test`.

### DIFF
--- a/test/suite/contextValue.test.ts
+++ b/test/suite/contextValue.test.ts
@@ -11,75 +11,87 @@ import { ContainerNode, FileNode, FolderNode, INodeData, NodeKind, PackageNode, 
 suite("Context Value Tests", () => {
 
     test("test workspace node", async function() {
-        assert.equal((await workspace.getTreeItem()).contextValue, "java:workspaceFolder+uri");
+        assert.ok(/java:workspaceFolder(?=.*?\b\+uri\b)/.test((await workspace.getTreeItem()).contextValue || ""));
     });
 
     test("test Maven project node", async function() {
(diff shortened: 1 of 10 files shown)


test/suite/contextValue.test.ts
```diff
@@ -11,75 +11,90 @@ import { ContainerNode, FileNode, FolderNode, INodeData, NodeKind, PackageNode,
...
     });
 
     test("test Maven container node", async function() {
-        assert.equal((await mavenContainer.getTreeItem()).contextValue, "java:container+maven+uri");
+        assert.ok(/java:container(?=.*?\b\+maven\b)(?=.*?\b\+uri\b)/.test((await mavenContainer.getTreeItem()).contextValue || ""));
     });
 
     test("test Gradle container node", async function() {
-        assert.equal((await gradleContainer.getTreeItem()).contextValue, "java:container+gradle+uri");
+        assert.ok(/java:container(?=.*?\b\+gradle\b)(?=.*?\b\+uri\b)/.test((await gradleContainer.getTreeItem()).contextValue || ""));
```
Eskibear, line 34: suggest to assign human-readable names to these regular expressions...

jdneo: These are all the regexp that can be written in the package.json to register the commands to the node.

I add some comments at header to explain the purpose of this suite.

Eskibear (approved):

PR #467, opened 2021-03-08
fix: Create new packages or classes does not work via hotkeys

### LINKED ISSUE #465: Creating new packages or classes doesn't work using hotkeys
image

Expected behaviour: User defined hotkeys allow creation of new classes or packages while focused on Java project explorer.

Actual behaviour: User defined hotkeys fail to launch "new package" and "new class" prompts. Clicking remains the only way to access these prompts.

jdneo:
This PR fix #465, including following changes

- When the commands `new package/class` are triggered and the parameter is not provided (triggered via hotkeys), get the current selected node.
- Since the selected node may not a valid node to create package/class, some condition checks are added.
- In addition, when creating class from a `Primary Type`, use its declared package as the base path.


src/explorerCommands/new.ts
```diff
@@ -90,7 +107,7 @@ function getNewFilePath(basePath: string, className: string): string {
 }
 
 export async function newPackage(node: DataNode): Promise<void> {
-    if (!node.uri) {
+    if (!node?.uri || !canCreatePackage(node)) {
```
Eskibear, line 110: Just one concern. If no node is selected, by triggering keyboard shortcut, nothing would happen. Is that true? If so, "Creating new packages or classes doesn't work using hotkeys" is still true in such scenarios.

I think it would be better to let users know what to do next. 
* Better experience: Prompt to select target package
* Easy implementation: notification with instructions

Eskibear (commented):
LGTM.

src/explorerCommands/new.ts
jdneo: Good point. Let me think how to improve this

Eskibear (approved):

PR #484, opened 2021-04-22
bugfix: Data tree node not found

### LINKED ISSUE #478: bug: Data tree node not found in console
After loading project view, an error/warning occurs in the debug console/developer tools console.
bug
bug1

jdneo:
fix #478

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -868,6 +868,11 @@
       "integrity": "sha512-Wm6ukoaOGJi/73p/cl2GvLjTI5JM1k/O14isD73YML8StrH/7/lRFgmg8nICZgD3bZZvjwCGxtMOD3wWNAu8cg==",
       "dev": true
     },
+    "await-lock": {
+      "version": "2.1.0",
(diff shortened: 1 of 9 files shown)


Eskibear:
It looks you are using `await-lock` to replace previously self-implemented one. So what's the root cause? Is previous one not working?

jdneo:
`await-lock` is just to reuse the module that has the same functionality.

The root cause of this bug is that, before this PR, there are different factors causing the explorer to refresh:
1. when reveal a node, it will trigger the explorer to `getChildren()` until the node is found
2. When server mode changed, the explorer will be refreshed entirely

Then problem happens, because the `refresh` is not an atomic operation (multiple `getChildren()` being called from different levels), thus the multiple `refresh` operation will race with each other, and break the cache.

So in this PR, I add some condition check that prevent those redundant `refresh` requests being called. Then during the activation, there will be only one `reveal` operation being called.

Eskibear (approved):

PR #498, opened 2021-05-19
fix: Multiple export jar terminals exist at the same time

### LINKED ISSUE #495: When exporting jar from explorer title, the export jar task is available
Currently the expect behavior of exporting jar is singleton and multiple exporting jar tasks are not supported at the same time. But when the exporting jar process comes from explorer's title, the export jar task is available.

CsCherrYY:
fix #495


CsCherrYY:
I will close this now. If there are any user complaints about #495 we may consider a solution for it.

PR #533, opened 2021-08-10
feat: Support new file contribution

jdneo:
1. If there is a active text editor and it's uri belongs to a workspace folder, then generate a new class at the same folder
2. Otherwise, ask user to select target project (if multiple) and package
3. If no project is opened, ask user to create a new project

resolve https://github.com/microsoft/vscode-java-pack/issues/706

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -262,9 +262,9 @@
       "dev": true
     },
     "@types/vscode": {
-      "version": "1.53.0",
-      "resolved": "https://devdiv.pkgs.visualstudio.com/_packaging/CloudKernel/npm/registry/@types/vscode/-/vscode-1.53.0.tgz",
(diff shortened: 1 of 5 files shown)


Eskibear:
LGTM, and suggest to support full qualified class name in future?

jdneo:
Change the logic to:

1. If current active editor is on the project source path, create a file under the directory of the active editor
2. Otherwise, create an untitled file with a simple new Java class snippet.

Eskibear (dismissed):

jdneo:
Appended a commit. Now if there is no active text editor but having one source path, the source path will be returned as the package path. This is useful in invisible projects.

jdneo:
Yes it's unstable

Eskibear (approved):

jdneo:
We might need to revisit this feature once we can get the selected file path from the file explorer. See: https://github.com/microsoft/vscode/issues/3553

PR #534, opened 2021-08-10
chore: Support tests in VSCode 1.59

CsCherrYY:


test/ui/index.ts
```diff
@@ -11,12 +11,12 @@ async function main(): Promise<void> {
         const testPath = path.join(__dirname, "command.test.js");
         const exTester = new ExTester();
         // The current version (4.1.1) of vscode-extension-tester doesn't support the newest VSCode version (^1.58.0)
-        await exTester.downloadCode("1.57.0");
+        await exTester.downloadCode("1.59.0");
```
jdneo, line 14: I think we should get the version number defined in package.json `"@types/vscode": "1.xx.x"`
`

test/ui/index.ts
```diff
@@ -11,12 +11,12 @@ async function main(): Promise<void> {
         const testPath = path.join(__dirname, "command.test.js");
         const exTester = new ExTester();
         // The current version (4.1.1) of vscode-extension-tester doesn't support the newest VSCode version (^1.58.0)
```
jdneo, line 13: This line can be removed

test/ui/command.test.ts
```diff
@@ -20,7 +20,7 @@ const targetPath = path.join(__dirname, "..", "..", "..", "test", "newProject");
 
 describe("Command Tests", function() {
 
-    this.timeout(60000);
+    this.timeout(120000);
```
jdneo, line 23: prefer tp writing as `2 * 60 * 1000 /*ms*/`

test/ui/index.ts
```diff
@@ -3,20 +3,26 @@
...
+import * as fse from "fs-extra";
+import * as semver from "semver";
 import { ExTester } from "vscode-extension-tester";
 
 async function main(): Promise<void> {
     try {
         // Run UI command tests
+        const packageContent = await fse.readFile(path.join(__dirname, "..", "..", "..", "package.json"));
+        const packageJSON = JSON.parse(packageContent.toString());
+        let vscodeVersion = packageJSON.engines.vscode || "1.59.0";
```
jdneo, line 15: No need to have `"1.59.0"`, if `packageJSON.engines.vscode` is unavailable, that indicates something is wrong in package.json, which is a good signal.

jdneo (approved):

PR #535, opened 2021-08-10
fix: Do not activate the Java Language Server

### LINKED ISSUE #532: Java Language Server get activated unexpectedly
Steps to repro:

1. Open a vs code window without any workspace folder opened.
2. Trigger the command `Create new Java Project`
3. The JLS get activated unexpectedly.

jdneo:
fix #532 

We should not activated JLS manually. Instead, polling it to check if it has been activated


Eskibear:
The polling model makes sense if you don't have any use case that you really want to activate JLS. Then it just works like a plugin.

jdneo:
Because now the project manager for java does not depend on JLS, so we should not activated JLS manually. (For example, triggering create new java project should not activate JLS)

Eskibear (approved):

PR #547, opened 2021-09-14
fix: Show the hints in Java Projects explorer properly

jdneo:
The PR is to let the Java Project explorer properly show hints for users:
- When `redhat.java` is not installed, hint for installation
- When no Java projects found, hint for creation/opening
- When JDT.LS in Lightweight mdoe, hint for switching.

Now the activation events is the same as `redhat.java`. But since events like `workspaceContains` is not supported in when clause, so I use APIs to query files during activation and set context value to imitate the condition:
- `resourceLangId == java` is for `onLanguage:java`
- `java:workspaceContainsBuildFiles` -> `workspaceContains`
- `java:serverMode` is to make sure the explorer will always visible once JDT.LS is activated. (still visible when user switch focus to a non-Java file)

### DIFF
--- a/src/languageServerApi/languageServerApiManager.ts
+++ b/src/languageServerApi/languageServerApiManager.ts
@@ -1,54 +1,58 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-import { commands, Event, Extension, extensions, Uri } from "vscode";
+import { commands, Disposable, Event, Extension, extensions, Uri } from "vscode";
 import { Commands } from "../commands";
+import { Context, ExtensionName } from "../constants";
 import { contextManager } from "../contextManager";
+import { Settings } from "../settings";
+import { syncHandler } from "../syncHandler";
 import { LanguageServerMode } from "./LanguageServerMode";
 
-class LanguageServerApiManager {
+class LanguageServerApiManager implements Disposable {
+    /**
+     * undefined means a legacy version language server
+     * null means the JDT.LS is not activated
+     */
     private serverMode: LanguageServerMode | null | undefined = null;
 
-    public async isStandardServerReady(): Promise<boolean> {
-        await this.checkServerMode();
-        // undefined serverMode indicates an older version language server
-        if (this.serverMode === undefined) {
-            return true;
-        }
+    private extensionChangeListener: Disposable;
 
-        if (this.serverMode !== LanguageServerMode.Standard) {
+    public async ready(): Promise<boolean> {
+        await this.checkServerMode();
+
+        if (this.serverMode === null || this.serverMode === LanguageServerMode.LightWeight) {
             return false;
         }
 
-        return true;
-    }
-
-    public async isLightWeightMode(): Promise<boolean> {
-        await this.checkServerMode();
-        return this.serverMode === LanguageServerMode.LightWeight;
-    }
-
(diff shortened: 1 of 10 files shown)


CsCherrYY:
After testing, I find no regression but an existing bug about export jar.

When there is no workspace and the language server has been activated, the export jar button doesn't show in the `Java Projects` view title. However, the user can trigger export jar via command palette `Java: Export Jar` and get the following error.

Screenshot 2021-09-14 170239

I'd like to make the `when` clause of `java.view.package.exportJar` in command palette be alignment with that in view title to fix this bug. @jdneo Could you please add a commit about this?

src/languageServerApi/languageServerApiManager.ts
```diff
@@ -1,94 +1,114 @@
...
+
+    public async ready(): Promise<boolean> {
         await this.checkServerMode();
-        // undefined serverMode indicates an older version language server
-        if (this.serverMode === undefined) {
-            return true;
-        }
 
-        if (this.serverMode !== LanguageServerMode.Standard) {
+        if (this.serverMode === null || this.serverMode === LanguageServerMode.LightWeight) {
```
Eskibear, line 24: what's the difference of this line? Or just clearer?

Eskibear: In which case, will this.serverMode be LanguageServerMode.Hybrid? On startup?

src/languageServerApi/languageServerApiManager.ts
```diff
@@ -1,94 +1,114 @@
...
-    public async isLightWeightMode(): Promise<boolean> {
-        await this.checkServerMode();
-        return this.serverMode === LanguageServerMode.LightWeight;
-    }
-
-    public async awaitSwitchingServerFinished(): Promise<void> {
-        await this.checkServerMode();
         if (this.serverMode === LanguageServerMode.Hybrid) {
             await new Promise<void>((resolve: () => void): void => {
                 extensions.getExtension("redhat.java")!.exports.onDidServerModeChange(resolve);
```
Eskibear, line 30: is it possible that serverMode changes from Hybrid to Lightweight?

src/languageServerApi/languageServerApiManager.ts
jdneo: When both lightweight server and standard server are working -> mostly standard server is activating, since once it's activated, lightweight server will be killed.

src/languageServerApi/languageServerApiManager.ts
jdneo: No, Currently we only support this state chain: `Lightweight` -> `Hybrid` -> `Standard`

Eskibear: Just curious what if I change settings from "hybrid" to "lightweight" during launching of standard LS? What's the result? will `serverMode` end up with "Standard" or "lightweight"?

jdneo: You will get standard - with a pop-up dialog to ask you reload window to make the setting change active

Eskibear: got it.

CsCherrYY (approved):

Eskibear (approved):
LGTM

CsCherrYY (approved):

PR #557, opened 2021-09-27
chore: Remove gulp dependencies

jdneo:


scripts/buildJdtlsExt.js
```diff
@@ -0,0 +1,37 @@
...
+    for (const jar of jars) {
+        fse.copyFileSync(path.join(sourceFolder, jar), path.join(targetFolder, path.basename(jar)));
+    }
+}
+
+function isWin() {
+    return /^win/.test(process.platform);
+}
+
+function isMac() {
```
CsCherrYY, line 27: `isMac()` and `isLinux()` are using in anywhere?

jdneo: They are legacy codes but I'm fine to remove them since they are not used anywhere

CsCherrYY (approved):

PR #569, opened 2021-12-06
Fix #568 type error

### LINKED ISSUE #568: Using `vscode-java-dependency` causes failures: Incorrect type. Expected one of object, boolean.
Hi there,
We use your extension in our product and because of a type issue in `package.json` we get issues in related processes,
Can you please fix it?
This is the issue in `package.json`:
image
 
Thanks

EstherPerelman:
Fixes #568

### DIFF
--- a/package.json
+++ b/package.json
@@ -568,6 +568,9 @@
             "items": {
               "anyOf": [
                 {
+                  "type": "string"
+                },
+                {
                   "enum": [
                     "${compileOutput}",
                     "${testCompileOutput}",
@@ -579,8 +582,7 @@
                     "%taskDefinitions.java.project.exportJar.testCompileOutput%",
                     "%taskDefinitions.java.project.exportJar.dependencies%",
                     "%taskDefinitions.java.project.exportJar.testDependencies%"
-                  ],
-                  "type":"string"
+                  ]
                 }
               ]
             },
--- a/package.json
+++ b/package.json
@@ -579,9 +579,9 @@
                     "%taskDefinitions.java.project.exportJar.testCompileOutput%",
                     "%taskDefinitions.java.project.exportJar.dependencies%",
                     "%taskDefinitions.java.project.exportJar.testDependencies%"
-                  ]
-                },
-                "string"
+                  ],
+                  "type":"string"
+                }
               ]
             },
             "description": "%taskDefinitions.java.project.exportJar.elements%"


jdneo:
@CsCherrYY

Could you help check if this update works?

CsCherrYY:
Here we use `anyOf` to accept both enum values and string values in export jar task definition. In this PR the string values will not be accepted.
image

@EstherPerelman You can reproduce is by `Terminal -> configure Tasks -> java: exportjar: xxx`, and the extension will create this template for you. BTW, I see the issue you mentioned is a warning, not an error, is it blocking you?

EstherPerelman:
> Here we use `anyOf` to accept both enum values and string values in export jar task definition. In this PR the string values will not be accepted.

@CsCherrYY thanks for replying, I updated the code to accept string value, It should work now

> BTW, I see the issue you mentioned is a warning, not an error, is it blocking you

Yes it is blocking me, we are taking the `taskDefinitions` contributions and use in an object and later fails on tasks processes because the object is unrecognize

CsCherrYY:
@EstherPerelman This patch works. Thanks for the help! 👍

CsCherrYY (approved):

EstherPerelman:
@CsCherrYY thank you for the quick merge😄, can you please release a patch ASAP?

CsCherrYY:
Actually, our plan is to release the new version about two weeks later. But if you want the new build right now, you can build a private bit yourself from the source code. Just check https://github.com/microsoft/vscode-java-dependency#contribution, then package a vsix file and install from VSIX file.

EstherPerelman:
Hi @CsCherrYY, 
There are [**SAP**](www.sap.com) customers who fall for this issue in production environments and **we must it ASAP**,

> you can build a private bit yourself from the source code
We have an automated process for the NPM that rely only on official releases from NPM so we can't build it...

Besides, **we are not interested in taking the whole new version** that will bring changes with it, we just want this fix (in a patch), we are now using version 0.18.1

jdneo:
> Besides, we are not interested in taking the whole new version that will bring changes with it, we just want this fix (in a patch), we are now using version 0.18.1

Little confused here. If we release a new version(Let's say `0.18.9`, which has quite large gap between `0.18.1`) with the fix here, will it work for you?

PR #575, opened 2021-12-22
chore: Deprecate document of `askUser`

CsCherrYY:
we now deprecate the documentation of `askUser` from configuration `java.project.exportJar.targetPath`, and the next step will be removing it totally.


package.json
```diff
@@ -210,13 +210,10 @@
...
               "enum": [
                 "${workspaceFolder}/${workspaceFolderBasename}.jar",
-                "askUser",
                 ""
               ],
               "enumDescriptions": [
-                "%configuration.java.project.exportJar.targetPath.workspaceFolder%",
-                "%configuration.java.project.exportJar.targetPath.askUser%",
-                "%configuration.java.project.exportJar.targetPath.askUser%"
+                "%configuration.java.project.exportJar.targetPath.workspaceFolder%"
```
jdneo, line 216: Add a description for `""` as well?

CsCherrYY: OK, added the description in https://github.com/microsoft/vscode-java-dependency/pull/575/commits/f1a42c2b5be3f895ab547cc9d24b8e206f220c5b

jdneo:
One more question, where can I see the description for the enums?

image

CsCherrYY:
> One more question, where can I see the description for the enums?
> 
> image

when you're editing this setting in raw settings.json

image

package.nls.json
```diff
@@ -28,11 +28,11 @@
   "configuration.java.dependency.packagePresentation": "Package presentation mode: flat or hierarchical",
   "configuration.java.project.exportJar.targetPath": "Specifies the output path of export jar. The default value is `${workspaceFolder}/${workspaceFolderBasename}.jar`. You can leave it empty to select the output location manually each time exporting the jar file.",
   "configuration.java.project.exportJar.targetPath.workspaceFolder": "Export the jar file into the workspace folder. Its name is the same as the folder's.",
-  "configuration.java.project.exportJar.targetPath.select": "You will select the output location manually each time exporting the jar file.",
+  "configuration.java.project.exportJar.targetPath.select": "Select output location manually when exporting the jar file.",
```
jdneo, line 31: how about:
```
"configuration.java.project.exportJar.targetPath.customization": "The output path of the exported jar. Leave it to empty if you want to manually pick the output location."
```

jdneo: My bad. It should be `leave it empty` instead of `leave it to empty`

jdneo (approved):

PR #592, opened 2022-02-10
chore: Update dependency

CsCherrYY:


CsCherrYY:
for globby > `12.0.0`, it has declarations like 
```
import fs from 'node:fs';
```

in its source. However, the schema `node:fs` has been supported in VS Code since node 16+, see https://github.com/microsoft/vscode/issues/123968#issuecomment-841900646 and https://github.com/microsoft/TypeScript/issues/43842#issuecomment-851018128 for more details.

And I just test in VSCode insiders, who is using node 16, where the error doesn't exist. So here I suggest updating 

- globby to 12.0.0+
- copy-webpack-plugin to 10.0.0+ (which requires globby 12+, see https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/CHANGELOG.md#1000-2021-11-17)

once the VS Code stable is using node 16.

jdneo:
So do we have a update solution for now?

CsCherrYY:
> So do we have a update solution for now?

actually, the mentioned dependencies have no CVE issues. So we can just keep waiting for the support of node 16 in VS Code stable.

jdneo (approved):

PR #602, opened 2022-03-17
Update dependency

CsCherrYY:


test/gradle/build.gradle
```diff
@@ -10,5 +10,5 @@ repositories {
 }
 
 dependencies {
-    compile 'com.google.code.gson:gson:2.8.0'
+    implementation 'com.google.code.gson:gson:2.8.0'
```
CsCherrYY, line 13: Since the current build-in Gradle version in buildship, which is integrated in Java language server, is `7.1.1`. This version has removed support of `compile`, we need to use `implementation`.

Now if you open this project you will see an error about `compile`.

jdneo (approved):

PR #603, opened 2022-03-24
prepare for pre-release channel

Eskibear:


Eskibear:
> src/views/dependencyExplorer.ts(101,42): error TS2345: Argument of type 'readonly ExplorerNode[]' is not assignable to parameter of type 'ExplorerNode[]'.

FYI, probably because of the engine update.

CsCherrYY (approved):
the change on the ts error works well.

PR #622, opened 2022-06-06
chore: Clean up the metadata files

jdneo:


jdneo:
Looks like some cases are failed after the metadata file is removed:

```
 1) Simple Project View Tests
       Can node render correctly:
	at done (/home/runner/work/vscode-java-dependency/vscode-java-dependency/node_modules/mocha/lib/mocha.js:1058:7)
AssertionError [ERR_ASSERTION]: Project name should be "1.helloworld"
+ expected- actual
	at processTicksAndRejections (node:internal/process/task_queues:96:5)
-simple_d4700ecf
+1.helloworld
```

will take a look.

CsCherrYY:
how about using some assertion like `assertTrue(project.getName().contains('simple'))` for invisible projects

jdneo:
> how about using some assertion like `assertTrue(project.getName().contains('simple'))` for invisible projects

That's a good idea. I just realized that maybe the original test is used to test Eclipse project? So I should not remove the metadata files?

CsCherrYY (approved):

PR #623, opened 2022-06-07
ux: Expose rebuild workspace command

### LINKED ISSUE #619: Provide a menu to rebuild the workspace
The rebuild operation should clean and build the workspace.

jdneo:
fix #619 
image

We will use the `tools` icon for now per discussion [here](https://github.com/microsoft/vscode-codicons/issues/111#issuecomment-1147646310)

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,8 @@
       },
       {
         "command": "java.project.build.workspace",
-        "title": "%contributes.commands.java.project.build.workspace%"
+        "title": "%contributes.commands.java.project.build.workspace%",
+        "icon": "$(tools)"
       },
       {
         "command": "java.project.clean.workspace",
@@ -398,7 +399,7 @@
           "group": "navigation@20"
         },
         {
-          "command": "java.view.package.refresh",
+          "command": "java.project.build.workspace",
           "when": "view == javaProjectExplorer && java:serverMode == Standard && !java:noJavaProjects",
           "group": "navigation@30"
         },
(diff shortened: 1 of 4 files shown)


CsCherrYY:
can we consider putting the rebuild button ahead of the "export jar"?

jdneo:
Any reason for that?

CsCherrYY:
IMO build/rebuild commands should have a higher priority than publishing artifact (export jar is under this kind) ones.

testforstephen (approved):

jdneo:
> IMO build/rebuild commands should have a higher priority than publishing artifact (export jar is under this kind) ones.

The priority of the icons in the navigation bar does not differ a lot. Since we only have 4 icons there (exclude the overflow button), put it as second or third position should not differ a lot.

From the UI perspective, put the build icon to the second one will look like this:
image
Personally, I don't like the visual effect because the icon of build is more complex than the ones next to it.

CsCherrYY:
> > IMO build/rebuild commands should have a higher priority than publishing artifact (export jar is under this kind) ones.
> 
> The priority of the icons in the navigation bar does not differ a lot. Since we only have 4 icons there (exclude the overflow button), put it as second or third position should not differ a lot.
> 
> From the UI perspective, put the build icon to the second one will look like this: image Personally, I don't like the visual effect because the icon of build is more complex than the ones next to it.

Get it and agree. Let's keep this version.

CsCherrYY (approved):

Issue #613, opened 2022-05-09
Support drag and drop in Project explorer


jdneo:
Below are the scenarios that we can support in the first iteration:

1. Drag java project explorer node to editor:
    - [x] primary type to editor
    - [x] document symbol to editor

2. DnD inside the java project explorer:
    - [x] Move file nodes (to folder, package, package root)
    - [x] Move folder nodes (to folder, package, package root)
    - [x] Move source types (to folder, package, package root)
    - [x] Move packages (to folder, package, package root)
    - [x] Move jars into Unmanaged Folder's Referenced Libraries node

3. DnD from VS Code's file explorer and the system's file explorer
    - [x] Move jars into Unmanaged Folder's Referenced Libraries node
    - [x] Copy files into Java Projects explorer from system file explorer

CsCherrYY:
tested in the newest pre-release (0.20.2022062500) and found an inconsistent.

![dnd](https://user-images.githubusercontent.com/45906942/176083062-95019008-6079-4c35-9cdc-192fef852698.gif)

can we also show a confirm dialog when moving members of a type?

jdneo:
> can we also show a confirm dialog when moving members of a type?

That's a bug I think. So far we should prevent moving members. I'll take a look.

PR #634, opened 2022-06-13
feat: Support drag and drop inside Java Project explorer

jdneo:
part of #613 

### DIFF
--- a/src/views/DragAndDropController.ts
+++ b/src/views/DragAndDropController.ts
@@ -29,7 +29,7 @@ export class DragAndDropController implements TreeDragAndDropController<Explorer
 
     public handleDrag(source: ExplorerNode[], treeDataTransfer: DataTransfer): void {
         // select many is not supported yet
-        let dragItem = source[0];
+        const dragItem = source[0];
         this.addDragToEditorDataTransfer(dragItem, treeDataTransfer);
         this.addInternalDragDataTransfer(dragItem, treeDataTransfer);
     }
@@ -126,7 +126,7 @@ export class DragAndDropController implements TreeDragAndDropController<Explorer
             return false;
         }
 
-       return this.isUnderSourceRoot(node);
+        return this.isUnderSourceRoot(node);
     }
 
     /**
--- a/src/views/DragAndDropController.ts
+++ b/src/views/DragAndDropController.ts
@@ -1,13 +1,22 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-import { DataTransfer, DataTransferItem, TreeDragAndDropController } from "vscode";
+import * as path from "path";
+import { commands, DataTransfer, DataTransferItem, TreeDragAndDropController, Uri, window, workspace, WorkspaceEdit } from "vscode";
+import { Commands } from "../commands";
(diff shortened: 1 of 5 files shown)


src/views/DragAndDropController.ts
```diff
@@ -37,4 +60,154 @@ export class DragAndDropController implements TreeDragAndDropController<Explorer
...
+    private async move(source: DataNode, target: DataNode): Promise<void> {
+        const sourceUri = Uri.parse(source.uri!);
+        const targetUri = Uri.parse(target.uri!);
+        if (sourceUri === targetUri) {
+            return;
+        }
+
+        const newPath = path.join(targetUri.fsPath, path.basename(sourceUri.fsPath));
+        const choice = await window.showInformationMessage(
+            `Are you sure you want to move '${path.basename(sourceUri.fsPath)}' into '${path.basename(targetUri.fsPath)}'?`,
```
CsCherrYY, line 201: When moving a class to another package, it seems not to change the package declaration, the same as dragging in the file explorer.

The Java language server supports move refactoring which will rename the package declaration and reference as well, is it possible we trigger this refactoring for class dragging to package here?

src/views/DragAndDropController.ts
```diff
@@ -37,4 +60,154 @@ export class DragAndDropController implements TreeDragAndDropController<Explorer
...
+            }
+            parent = parent.getParent();
+        }
+        return true;
+    }
+
+    /**
+     * Check whether the node is able to be dropped.
+     */
+    private isDroppableNode(node: ExplorerNode | undefined): boolean {
```
CsCherrYY, line 157: It's a little strange for me to drag a resource folder to a Java package: After I confirm, I can't find that folder in the destination package (since we will only show classes and packages under a package).

src/views/DragAndDropController.ts
jdneo: That's a regression bug in 1.7.0: https://github.com/redhat-developer/vscode-java/issues/2503

you can build a private build with the latest code or wait for the next pre-release. 😄

src/views/DragAndDropController.ts
jdneo: I think that folder will become a child package of the dropped package. What's the behavior on your side?

CsCherrYY: the package node shows green (new files exist) but the dragged files can't be found.

dnd

jdneo: I found that in this case, Eclipse can show `static` as a normal folder, while we cannot.

The root cause is here: https://github.com/microsoft/vscode-java-dependency/blob/2a6c0fc9fa07480dd28d4f1f7214a946f9136f17/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java#L406

We didn't take `IFolder` into consideration and just simply skip it.

I think we can still support moving resource folders, and fix the bug to show folders in packages.

jdneo: Filed an issue for this: https://github.com/microsoft/vscode-java-dependency/issues/635

CsCherrYY (approved):

PR #636, opened 2022-06-14
fix: Add overwrite confirmation

jdneo:
part of #613 

This PR adds a overwrite confirmation when the target file already exists in the file system.

### DIFF
--- a/src/views/DragAndDropController.ts
+++ b/src/views/DragAndDropController.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 import * as path from "path";
+import * as fse from "fs-extra";
 import { commands, DataTransfer, DataTransferItem, TreeDragAndDropController, Uri, window, workspace, WorkspaceEdit } from "vscode";
 import { Commands } from "../commands";
 import { Explorer } from "../constants";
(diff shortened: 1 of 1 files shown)


src/views/DragAndDropController.ts
```diff
@@ -203,11 +204,32 @@ export class DragAndDropController implements TreeDragAndDropController<Explorer
...
+
+    /**
+     * Confirm the overwrite action when the target file already exists.
+     * @param file the path of the target file.
+     */
+    private async confirmOverwrite(file: string): Promise<boolean> {
+        if (await fse.pathExists(file)) {
+            const name = path.basename(file);
+            const ans = await window.showWarningMessage(
+                `A file or folder with the name '${name}' already exists in the destination folder. Do you want to replace it?`,
```
CsCherrYY, line 223: if we're going to move a folder, and there is a folder with the same name in the dest folder, what will happen? merge the contents or just replace?

jdneo: It will replace - same behavior as the file explorer.

CsCherrYY: makes sense

CsCherrYY (approved):

PR #637, opened 2022-06-15
feat: Support adding referenced libraries via drag and drop

jdneo:
part of #613. 

Drag folders from the VS Code's file explorer to the referenced libraries node is not supported due to an issue: https://github.com/microsoft/vscode/issues/152031


CsCherrYY (approved):
tested in an unmanaged folder, it works well.

PR #651, opened 2022-06-28
fix: Document symbol should not be dropped to Project explorer

jdneo:
fix https://github.com/microsoft/vscode-java-dependency/issues/613#issuecomment-1168162227


jdneo:
CI for MacOS fails. It seems like related with https://github.com/microsoft/vscode-test/issues/153

CsCherrYY (approved):
LGTM.

> CI for MacOS fails. It seems like related with https://github.com/microsoft/vscode-test/issues/153

I'll create a new PR to update the test framework to the newest `@vscode/test-electron` these days.

PR #660, opened 2022-07-07
feat: Support run build from tasks

jdneo:
- Add a new task provider with type: 'java (build)'.
- Use 'paths' to define the root path of the project to be built.
- Use 'isFullBuild' to define whether this build is a full built or not.


src/tasks/build/buildTaskProvider.ts
```diff
@@ -0,0 +1,268 @@
...
+        if (includedPaths.length === 0 || this.cancellationTokenSource.token.isCancellationRequested) {
+            return;
+        }
+
+        this.printList("Building following projects:", includedPaths);
+        const uris: Uri[] = includedPaths.map(p => Uri.file(p));
+        try {
+            const res = await commands.executeCommand(Commands.BUILD_PROJECT, uris, this.definition.isFullBuild,
+                this.cancellationTokenSource.token);
+            if (res === Jdtls.CompileWorkspaceStatus.Witherror && checkErrorsReportedByJavaExtension()) {
```
CsCherrYY, line 147: I suggest to use `closeEmitter.fire()` to indicate the task result, in this way the user can easily distinguish task result in the terminal  UI. e.g, https://github.com/microsoft/vscode-gradle/blob/a9d5977d0a8505ddc5bf55000046e0e96428e4d3/extension/src/terminal/GradleRunnerTerminal.ts#L105-L108

jdneo: Do you mean call `closeEmitter.fire()` in the finally block?

CsCherrYY: currently we always use fire(0) (see: https://github.com/microsoft/vscode-java-dependency/blob/2759b924f89347439f5b31a77130845c764cfb14/src/tasks/build/buildTaskProvider.ts#L97) and it will return a successful task state, if we fire another value, e.g., fire(1) for failed condition, the failed state can be shown.

jdneo: Got your point, good suggestion!

CsCherrYY (approved):
LGTM

PR #662, opened 2022-07-12
chore: Update `globby` and `copy-webpack-plugin`

CsCherrYY:
it's a follow-up of https://github.com/microsoft/vscode-java-dependency/pull/592#issuecomment-1035902532

### DIFF
--- a/package.json
+++ b/package.json
@@ -650,7 +650,7 @@
     "@types/semver": "^7.3.10",
     "@types/vscode": "1.69.0",
     "@vscode/test-electron": "^2.1.5",
-    "copy-webpack-plugin": "^9.1.0",
+    "copy-webpack-plugin": "^11.0.0",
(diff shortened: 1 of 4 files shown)


jdneo:
Should also update the `@types/node` to 16 I guess

package.json
```diff
@@ -646,11 +646,11 @@
     "@types/lodash": "^4.14.182",
     "@types/minimatch": "^3.0.3",
     "@types/mocha": "^9.1.1",
-    "@types/node": "^14.18.21",
+    "@types/node": "16.11.7",
```
jdneo, line 649: Or maybe just `16.x`. See: https://github.com/microsoft/vscode/blob/47a1f80c6531121a8ef450a980451f96a4a7e047/package.json#L111

jdneo (approved):

PR #663, opened 2022-07-13
feat: Add build project command into context menu

jdneo:
The build project command will be put into the same group as the run/debug commands

image


testforstephen (approved):
LGTM

PR #667, opened 2022-07-19
UX: Change export task type to `java (buildArtifact)`

### LINKED ISSUE #665: Rename the type of export jar task
Suggest renaming it from `java` to `java (export)`.

Besides that, we should reserve a task resolver to keep supporting the old type. And track the usage of the old type task.

CsCherrYY:
fix #665 

This pr also renames the default label from `java: exportjar:${workspaceName}` to `java (buildArtifact): ${workspaceName}`. We can still find `java` type via `Run Tasks...`, but the provider will not provide any template tasks anymore. If there is no `java` tasks in your tasks.json, the result of clicking `java` type will be empty.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -508,7 +508,7 @@
         {
           "command": "java.project.build.workspace",
           "when": "view == javaProjectExplorer && viewItem =~ /java:project(?=.*?\\b\\+java\\b)(?=.*?\\b\\+uri\\b)/",
-          "group":"8_execution@5"
+          "group": "8_execution@5"
         },
         {
           "command": "java.project.rebuild",
@@ -635,7 +635,7 @@
     ],
     "taskDefinitions": [
       {
-        "type": "java",
+        "type": "java (buildArtifact)",
         "properties": {
           "label": {
             "type": "string",
@@ -722,6 +722,64 @@
             "type": "boolean",
             "default": "true",
             "description": "%taskDefinitions.java.project.build.isFullBuild%"
+          }
+        }
+      },
+      {
+        "type": "java",
+        "when": "java:showDeprecatedTasks",
+        "properties": {
+          "label": {
+            "type": "string",
+            "description": "%taskDefinitions.java.project.exportJar.label%"
+          },
+          "mainClass": {
+            "type": "string",
+            "description": "%taskDefinitions.java.project.exportJar.mainClass%"
+          },
+          "targetPath": {
+            "type": "string",
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "enum": [
+                  "${workspaceFolder}/${workspaceFolderBasename}.jar",
+                  ""
+                ],
+                "enumDescriptions": [
+                  "%configuration.java.project.exportJar.targetPath.workspaceFolder%",
+                  "%configuration.java.project.exportJar.targetPath.select%"
+                ]
+              }
+            ],
+            "description": "%configuration.java.project.exportJar.targetPath.customization%"
+          },
+          "elements": {
+            "type": "array",
+            "items": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "enum": [
+                    "${compileOutput}",
+                    "${testCompileOutput}",
+                    "${dependencies}",
+                    "${testDependencies}"
+                  ],
+                  "enumDescriptions": [
+                    "%taskDefinitions.java.project.exportJar.compileOutput%",
+                    "%taskDefinitions.java.project.exportJar.testCompileOutput%",
+                    "%taskDefinitions.java.project.exportJar.dependencies%",
+                    "%taskDefinitions.java.project.exportJar.testDependencies%"
+                  ]
+                }
+              ]
+            },
+            "description": "%taskDefinitions.java.project.exportJar.elements%"
           }
         }
       }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-
 import * as path from "path";
-import { commands, Diagnostic, Extension, ExtensionContext, extensions, languages, tasks, TextDocument, TextEditor, Uri, window, workspace } from "vscode";
+import { commands, Diagnostic, Extension, ExtensionContext, extensions, languages,
+    Range, tasks, TextDocument, TextEditor, Uri, window, workspace } from "vscode";
 import { dispose as disposeTelemetryWrapper, initializeFromJsonFile, instrumentOperation, instrumentOperationAsVsCodeCommand, sendInfo } from "vscode-extension-telemetry-wrapper";
 import { Commands, contextManager } from "../extension.bundle";
 import { BuildTaskProvider } from "./tasks/build/buildTaskProvider";
@@ -11,11 +11,14 @@ import { buildFiles, Context, ExtensionName } from "./constants";
 import { LibraryController } from "./controllers/libraryController";
 import { ProjectController } from "./controllers/projectController";
 import { init as initExpService } from "./ExperimentationService";
-import { ExportJarTaskProvider } from "./exportJarSteps/ExportJarTaskProvider";
+import { DeprecatedExportJarTaskProvider, ExportJarTaskProvider } from "./exportJarSteps/ExportJarTaskProvider";
 import { Settings } from "./settings";
 import { syncHandler } from "./syncHandler";
 import { EventCounter } from "./utility";
 import { DependencyExplorer } from "./views/dependencyExplorer";
+import { DiagnosticProvider } from "./tasks/migration/DiagnosticProvider";
+import { setContextForDeprecatedTasks, updateExportTaskType } from "./tasks/migration/utils";
+import { CodeActionProvider } from "./tasks/migration/CodeActionProvider";
 
 export async function activate(context: ExtensionContext): Promise<void> {
     contextManager.initialize(context);
@@ -40,6 +43,7 @@ async function activateExtension(_operationId: string, context: ExtensionContext
     context.subscriptions.push(DependencyExplorer.getInstance(context));
     context.subscriptions.push(contextManager);
     context.subscriptions.push(syncHandler);
+    context.subscriptions.push(tasks.registerTaskProvider(DeprecatedExportJarTaskProvider.type, new DeprecatedExportJarTaskProvider()));
     context.subscriptions.push(tasks.registerTaskProvider(ExportJarTaskProvider.exportJarType, new ExportJarTaskProvider()));
     context.subscriptions.push(tasks.registerTaskProvider(BuildTaskProvider.type, new BuildTaskProvider()));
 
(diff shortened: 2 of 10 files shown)


CsCherrYY:
for workspace doesn't contain "java" tasks:

https://user-images.githubusercontent.com/45906942/179884446-20fa5a24-4804-4595-bf50-2edf6676bcae.mp4

for workspace contains "java" tasks:

https://user-images.githubusercontent.com/45906942/179884466-197746a6-260c-42d5-8e87-2e5d83197bc9.mp4

src/extension.ts
```diff
@@ -61,3 +65,25 @@ function addExtensionChangeListener(context: ExtensionContext): void {
...
+async function handleDeprecatedTasks(): Promise<void> {
+    await contextManager.setContextValue(Context.SHOW_DEPRECATED_TASKS, true);
+    const deprecatedTasks = await tasks.fetchTasks({ type: "java" });
+    if (deprecatedTasks?.length) {
+        let tasksJsonPath;
+        const options = [];
+        if (workspace.workspaceFolders?.length === 1) {
+            tasksJsonPath = path.join(workspace.workspaceFolders[0].uri.fsPath, ".vscode", "tasks.json");
+            if (await fse.pathExists(tasksJsonPath)) {
+                options.push("Open tasks.json");
```
jdneo, line 78: I prefer to providing an option that directly update the task type (if we can find the correct deprecated tasks).

CsCherrYY:
now we will show quick fix as well:

https://user-images.githubusercontent.com/45906942/180373429-725ea7a9-9e7f-4d5b-8eac-db2c17bc66a1.mp4

src/tasks/migration/CodeActionProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    constructor() {
+        vscode.commands.registerCommand(Commands.JAVA_UPDATE_DEPRECATED_TASK, async (filePath) => {
+            await updateExportTaskType(filePath);
+        });
+    }
+
+    public provideCodeActions(document: vscode.TextDocument, range: vscode.Range | vscode.Selection,
+                              _context: vscode.CodeActionContext, _token: vscode.CancellationToken): vscode.CodeAction[] | undefined {
+        const diagnostics = diagnosticProvider.getDiagnosticCollection().get(document.uri);
```
jdneo, line 22: We can use `languages.getDiagnostics(document.uri);`, by doing this, we can avoid writing `export const diagnosticProvider = new DiagnosticProvider();` in `DiagnosticProvider.ts`.

// I'm not a fan of export global variable like that since it will introduce a lot of problems when we write unit tests (with webpack enabled).

package.json
```diff
@@ -179,6 +179,11 @@
         "command": "java.view.package.renameFile",
         "title": "%contributes.commands.java.view.package.renameFile%",
         "category": "Java"
+      },
+      {
+        "command": "java.updateDeprecatedTask",
```
jdneo, line 184: Will the command work if we do not register it in `package.json`? I guess it should be.

src/tasks/migration/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+            const tasksJsonPath = path.join(folder.uri.fsPath, ".vscode", "tasks.json");
+            if (await fse.pathExists(tasksJsonPath)) {
+                tasksJsonPaths.push(tasksJsonPath);
+            }
+        }
+    }
+    return tasksJsonPaths;
+}
+
+export function asyncDebounce(func: any, wait: any, bind: any) {
```
jdneo, line 27: Why we need to wrapper it to async?

src/tasks/migration/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    if (taskJSON) {
+        const taskArray = taskJSON.tasks;
+        if (taskArray instanceof Array) {
+            for (const task of taskArray) {
+                if (task.type === DeprecatedExportJarTaskProvider.type) {
+                    task.type = ExportJarTaskProvider.exportJarType;
+                }
+            }
+        }
+        await fse.writeFile(tasksJsonPath, JSON.stringify(taskJSON, null, "\t"));
```
jdneo, line 58: We should get the EOL from the document instance.

Or, maybe we can just simply do a string replacement via regex matching.

src/tasks/migration/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+export async function handleDeprecatedTasks(): Promise<void> {
```
jdneo, line 62: Is this function being used anywhere?

src/tasks/migration/CodeActionProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+                    kind: vscode.CodeActionKind.QuickFix,
+                    command: updateTaskCommand
+                }];
+            }
+        }
+        return [];
+    }
+}
+
+export const codeActionProvider: CodeActionProvider = new CodeActionProvider();
```
jdneo, line 49: Same here, avoid export a global variable if it's not necessary.

src/tasks/migration/DiagnosticProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+    private static async getDiagnosticsFromTasksJsonPath(tasksJsonPath: string): Promise<vscode.Diagnostic[]> {
+        const diagnostics: vscode.Diagnostic[] = [];
+        const fileStream = fs.createReadStream(tasksJsonPath);
+        let lineNumber = 0;
+        const rl = readline.createInterface({
+            input: fileStream,
+            crlfDelay: Infinity
+        });
+        for await (const line of rl) {
+            const columnNumber = line.indexOf(DiagnosticProvider.DEPRECATED_TASK_TYPE_DECLARATION);
```
jdneo, line 59: What if there is no space between `": "`?

Looks like a safer way is to parse the json.

package.json
CsCherrYY: It works, let's remove the command registration.

src/tasks/migration/utils.ts
CsCherrYY: we check the status in `activateExtension()` when activating. it will pop up a notification if deprecated tasks detected.

jdneo: Oh yes, I missed it.

jdneo: I now have a feeling that, pops up such notifications is not a good UX. I think we should:

- Provide option like `Don't ask again`
- Only show it when user opens a `tasks.json` file

**Or maybe even just don't pop it, since we have diagnostics.**

It just like when we deprecate a setting, VS Code will only show diagnostics and will not pop notifications for them.

src/tasks/migration/DiagnosticProvider.ts
CsCherrYY: I didn't find a way to get the exact row and column when parsing the json, so here just read the file line by line.

src/tasks/migration/utils.ts
CsCherrYY: > We should get the EOL & indent from the document instance.

Yeah we should consider them.

> Or, maybe we can just simply do a string replacement via regex matching.

if we use regex matching, how can we handle if there is the same value in a string constant? (indeed, it's a very rare scenario) IMO it better to keep JSON's schema.

src/tasks/migration/utils.ts
CsCherrYY: it seems async function is not proper debounced from lodash debounce library. here I use the workaround here: https://github.com/lodash/lodash/issues/4815#issuecomment-859379489

jdneo: But we are not awaiting the function returned value. I think we can just kick off the execution without awaiting it. Do I miss anything here?

src/tasks/migration/DiagnosticProvider.ts
jdneo: What about regex match

src/tasks/migration/utils.ts
jdneo: Maybe you can deal it as a string, replace the part that needs to be changed (the diagnostic contains range information which could be useful), and write it back - then the indent will not be broken.

src/tasks/migration/utils.ts
CsCherrYY: get it. we can directly use debounce here.

src/extension.ts
```diff
@@ -38,8 +41,21 @@ async function activateExtension(_operationId: string, context: ExtensionContext
...
+    // handle deprecated tasks
+    const diagnosticProvider = new DiagnosticProvider();
+    await diagnosticProvider.initialize();
+    context.subscriptions.push(diagnosticProvider),
+    await setContextForDeprecatedTasks();
+    context.subscriptions.push(languages.registerCodeActionsProvider([{
+        scheme: "file",
+        pattern: "**/.vscode/tasks.json"
+    }], new CodeActionProvider()));
+    commands.registerCommand(Commands.JAVA_UPDATE_DEPRECATED_TASK, async (document: TextDocument, range: Range) => {
```
jdneo, line 56: Instrument it.

src/tasks/migration/CodeActionProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+import { Commands } from "../../commands";
+import { ExportJarTaskProvider } from "../../exportJarSteps/ExportJarTaskProvider";
+import { DiagnosticProvider } from "./DiagnosticProvider";
+
+export class CodeActionProvider implements vscode.CodeActionProvider {
+
+    public static JAVA_UPDATE_DEPRECATED_TASK_TITLE = "Update deprecated tasks";
```
jdneo, line 11: Looks like it's not used anywhere

src/tasks/migration/DiagnosticProvider.ts
```diff
@@ -0,0 +1,81 @@
...
+    private disposables: vscode.Disposable[] = [];
+    private refreshDiagnosticsTrigger: any;
+
+    public dispose() {
+        for (const d of this.disposables) {
+            d.dispose();
+        }
+    }
+
+    public async initialize(): Promise<void> {
```
jdneo, line 25: Can we just put the initialization logic into constructor and not await `initializeDiagnostics()`?

src/tasks/migration/DiagnosticProvider.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.refreshDiagnosticsTrigger = _.debounce(this.refreshDiagnostics, 500 /** ms */);
+        this.diagnosticCollection = vscode.languages.createDiagnosticCollection("migrateExportTask");
+        this.disposables.push(this.diagnosticCollection);
+        this.disposables.push(vscode.workspace.onDidChangeTextDocument(async (e) => {
+            if (e.document.fileName.includes("tasks.json")) {
```
jdneo, line 30: path.basename(xxx) === "tasks.json"

src/extension.ts
```diff
@@ -45,17 +45,17 @@ async function activateExtension(_operationId: string, context: ExtensionContext
     context.subscriptions.push(tasks.registerTaskProvider(ExportJarTaskProvider.exportJarType, new ExportJarTaskProvider()));
     context.subscriptions.push(tasks.registerTaskProvider(BuildTaskProvider.type, new BuildTaskProvider()));
     // handle deprecated tasks
-    const diagnosticProvider = new DiagnosticProvider();
-    await diagnosticProvider.initialize();
-    context.subscriptions.push(diagnosticProvider),
+    context.subscriptions.push(new DiagnosticProvider());
     await setContextForDeprecatedTasks();
```
jdneo, line 49: Let's move this line to the end of the function and remove the `await`.

jdneo (approved):

PR #671, opened 2022-07-21
feat: Show reload button at editor title when there is reload required diagnostics

jdneo:

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -64,6 +64,8 @@ export namespace Commands {
 
     export const JAVA_PROJECT_UPDATE = "java.project.update";
 
+    export const JAVA_PROJECT_RELOAD_ALL = "java.project.reloadProjectFromActiveFile";
+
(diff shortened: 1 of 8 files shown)


src/commands.ts
```diff
@@ -64,6 +64,8 @@ export namespace Commands {
 
     export const JAVA_PROJECT_UPDATE = "java.project.update";
 
+    export const JAVA_PROJECT_RELOAD_ALL = "java.project.reloadProjectFromActiveFile";
```
CsCherrYY, line 67: `JAVA_PROJECT_RELOAD_ALL` seems to reload all the projects, but it seems to update only the active build file's project in the command execution. Which is right?

jdneo: My bad, should be `JAVA_PROJECT_RELOAD_ACTIVE_FILE`

CsCherrYY (approved):
LGTM

PR #678, opened 2022-07-27
refactor: Move export jar related code under tasks folder

jdneo:


CsCherrYY (approved):
Should we update the wiki as well?

jdneo:
Yes, wiki should be updated, and the VS Code Java document.

PR #684, opened 2022-08-02
fix: Remove the limitation of only displaying invisible projects

### LINKED ISSUE #654: Projects which have linked resources is not supported well.
I'm currently trying to integrate https://github.com/jbangdev/jbang-eclipse with jdt.ls, which kinda worsS, but I noticed the Java Projects view doesn't show the JBang classpath container defined in https://github.com/jbangdev/jbang-eclipse/blob/main/dev.jbang.eclipse.core/plugin.xml#L38. In Eclipse it looks like: 

The label is defined in https://github.com/jbangdev/jbang-eclipse/blob/main/dev.jbang.eclipse.core/src/main/java/dev/jbang/eclipse/core/internal/classpath/JBangClasspathContainer.java#L26, but I don't know if there's an easy way to get to that, using JDT's API.

In any case, not showing the dependencies at all is very annoying.

Screenshot 2022-06-28 at 20 05 43

jdneo:
- now, no matter where the projects are located, we will display all of
  them in Java Projects explorer. (The only exception is the default
  project).

fix #654 


jdneo:
@fbricon

image

The next step is to make sure the `Referenced Libraries` will only appear for unmanaged folder, which requires to adding a new nature id for the unmanaged folders.

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
```diff
@@ -94,24 +94,25 @@ private static class Classpath {
...
+            if (Objects.equals(project.getName(), ProjectsManager.DEFAULT_PROJECT_NAME)) {
                 continue;
             }
+
             PackageNode projectNode = PackageNode.createNodeForProject(JavaCore.create(project));
-            if (Objects.equals(project.getName(), invisibleProjectName)) {
+
+            // set the folder name as the project name when the project location
+            // is out of the workspace folder.
+            if (!workspaceFolderPath.isPrefixOf(project.getLocation())) {
```
CsCherrYY, line 115: is it possible that there are multiple projects out of the workspace folder? In this scenario, they will have the same name.

jdneo: Yes, that could be, though not very likely to happen. We can consider showing the path of the node in the tooltip, like the behavior in file explorer, if we see users complaining about this.

CsCherrYY (approved):

PR #688, opened 2022-09-07
chore(deps): Update dependencies

jdneo:

### DIFF
--- a/package.json
+++ b/package.json
@@ -803,11 +803,11 @@
   "devDependencies": {
     "@types/fs-extra": "^9.0.13",
     "@types/glob": "^7.2.0",
-    "@types/lodash": "^4.14.182",
+    "@types/lodash": "^4.14.184",
     "@types/minimatch": "^3.0.3",
     "@types/mocha": "^9.1.1",
     "@types/node": "16.x",
-    "@types/semver": "^7.3.10",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -803,20 +803,20 @@
...
     "@types/vscode": "1.69.0",
     "@vscode/test-electron": "^2.1.5",
     "copy-webpack-plugin": "^11.0.0",
     "glob": "^7.2.3",
     "mocha": "^9.2.2",
     "ts-loader": "^9.3.1",
     "tslint": "^6.1.3",
-    "typescript": "^4.7.4",
-    "vscode-extension-tester": "^4.3.0",
+    "typescript": "^4.9.0-dev.20220907",
```
CsCherrYY, line 818: should we use dev version?

jdneo: I have no idea why but when I use latest stable version, it will throw OOM error when doing webpack.

And on the other hand, this is the version that VS Code itself used as well: https://github.com/microsoft/vscode/blob/6c6518c10a561bc6eccaa1b565280f63f9746c13/package.json#L209

CsCherrYY: get it, let's use this one

CsCherrYY (approved):

PR #697, opened 2022-10-31
chore: Prepare 0.21.1

jdneo:


jdtls.ext/com.microsoft.jdtls.ext.core/.project
```diff
@@ -33,12 +33,12 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1655120454990</id>
+			<id>1665551209562</id>
 			<name></name>
 			<type>30</type>
 			<matcher>
 				<id>org.eclipse.core.resources.regexFilterMatcher</id>
-				<arguments>node_modules|.git|__CREATED_BY_JAVA_LANGUAGE_SERVER__</arguments>
```
CsCherrYY, line 41: should we also check in this change?

jdneo: I think it's fine

CsCherrYY (approved):

PR #701, opened 2022-11-11
fix: Add eol after emitting export jar end message

### LINKED ISSUE #699: Output of the exporting jar task can be improved
image

create a new line at the end of the task

CsCherrYY:
fix #699

### DIFF
--- a/src/tasks/buildArtifact/BuildArtifactTaskProvider.ts
+++ b/src/tasks/buildArtifact/BuildArtifactTaskProvider.ts
@@ -206,7 +206,7 @@ class ExportJarTaskTerminal implements Pseudoterminal {
 
     public exit(message?: string) {
         if (message) {
-            this.writeEmitter.fire(message);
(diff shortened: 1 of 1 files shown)


src/tasks/buildArtifact/BuildArtifactTaskProvider.ts
```diff
@@ -206,7 +206,7 @@ class ExportJarTaskTerminal implements Pseudoterminal {
 
     public exit(message?: string) {
         if (message) {
-            this.writeEmitter.fire(message);
+            this.writeEmitter.fire(message + "\r\n");
```
jdneo, line 209: If `message` is undefined, will the output look good?

CsCherrYY: if message is undefined, we will not invoke writeEmitter so nothing will happen in the outout panel.

jdneo (approved):

PR #730, opened 2023-04-04
feat: Display non-java resources in the Project explorer

### LINKED ISSUE #145: Display non-Java files in Java Projects explorer
Eclipse has a very good structure for Java Projects . 

It shows pom.xml file of project and the other outer project folders . 

eclipse_2019-02-28_16-35-08

Here i can't find those  and all the time i am going between explorer and Java Dependencies Viewer : 
code_2019-02-28_16-31-32

Like let's add also everything that Explorer contains ... like the outer folders and showing pom.xml etc .. look how beautiful is in Eclipse ( one for all , not going back and forth with two viewers like here...)

And finally ... for god mercy why we can't create new file or delete like we can do in Explorer ? Come on .... :) Make me love vs code more by adding those please :)

jdneo:
Fix #145 

The new explorer will looks like this:
image

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
@@ -320,7 +320,6 @@ public class PackageCommand {
 
                 for (IPackageFragmentRoot fragmentRoot : packageFragmentRoots) {
                     children.add(fragmentRoot);
-                    children.addAll(Arrays.asList(fragmentRoot.getNonJavaResources()));
                 }
             }
         } catch (CoreException e) {
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
@@ -78,10 +78,10 @@ public class PackageCommand {
 
     static {
         commands = new HashMap<>();
-        commands.put(NodeKind.PROJECT, PackageCommand::getContainers);
-        commands.put(NodeKind.CONTAINER, PackageCommand::getPackageFragmentRoots);
-        commands.put(NodeKind.PACKAGEROOT, PackageCommand::getPackages);
-        commands.put(NodeKind.PACKAGE, PackageCommand::getRootTypes);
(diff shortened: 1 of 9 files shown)


jdneo:
TODO:

1.  Align the UI
2. Fix tests
3. File watcher update

jdneo:
#### Performance impact

##### Test case

After project petclinic imported, Open the file `PetClinicIntegrationTests.java` to let Explorer fetch the nodes and sync the position.

Without this PR:

| request | total time |
|---|---|
| java.resolvePath | 566 |
| java.getPackageData | 69 |

With this PR:

| request | total time |
|---|---|
| java.resolvePath | 354 |
| java.getPackageData | 85 |

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
```diff
@@ -311,6 +320,7 @@ private static List<PackageNode> getPackageFragmentRoots(PackageParams query, IP
 
                 for (IPackageFragmentRoot fragmentRoot : packageFragmentRoots) {
                     children.add(fragmentRoot);
+                    children.addAll(Arrays.asList(fragmentRoot.getNonJavaResources()));
```
CsCherrYY, line 323: the `nonJavaResources` should be under the fragment root node or do some filtering? otherwise we will get a lot of META-INFs in the Project and Dependencies node.

image

jdneo: Good finding!

I should not call `getNonJavaResources()` here. Here we just need to get all the jar fragments as the children of the container node. Those `META-INF` folders belong to the jars and should be shown later when user expand the jar node.

CsCherrYY:
Here is a case about non-Java project, given Gradle sample project (https://docs.gradle.org/current/samples/sample_building_java_applications.html), the current Java projects view shows only project `app` since it's a java project, and ignore the root project `demo` since it's a non-java project. 

Since we have supported displaying non-java resources, we can also consider showing the non-java project in the project explorer. Given eclipse package explorer:
image

jdneo:
> Here is a case about non-Java project, given Gradle sample project

Filed an issue to track: https://github.com/microsoft/vscode-java-dependency/issues/736

CsCherrYY (commented):
LGTM overall. One small thing: can we adjust the order of source sets? Showing main/test source sets before the generated source sets would be better. 

after: 
image

before:
image

CsCherrYY (approved):

PR #731, opened 2023-04-04
refactor: Use factory pattern to create explorer nodes

jdneo:
- Use factory pattern to create nodes. This can remove a lot of duplicated codes.
- Redefine the `NodeKind` enum to better sort the nodes in explorer

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
@@ -367,17 +367,20 @@ public class PackageCommand {
             return packageRoot;
         }
 
-        // jar in Referenced Libraries must be constructed from path
-        IJavaProject javaProject = getJavaProject(query.getProjectUri());
-        if (javaProject != null) {
-            try {
-                packageRoot = javaProject.findPackageFragmentRoot(Path.fromPortableString(query.getRootPath()));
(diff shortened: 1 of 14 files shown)


jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
```diff
@@ -367,17 +367,20 @@ private static IPackageFragmentRoot getPackageFragmentRootFromQuery(PackageParam
...
         }
 
-        // jar in Referenced Libraries must be constructed from path
-        IJavaProject javaProject = getJavaProject(query.getProjectUri());
-        if (javaProject != null) {
-            try {
-                packageRoot = javaProject.findPackageFragmentRoot(Path.fromPortableString(query.getRootPath()));
-            } catch (JavaModelException e) {
-                return null;
+        if (query.getProjectUri() != null && query.getRootPath() != null) {
```
Eskibear, line 370: just to check, is `query.getProjectUri()` cheap? if not, suggest to assign it to a var.

Eskibear (approved):
much more clear

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/PackageCommand.java
jdneo: Yes, they are just simple getters.

PR #747, opened 2023-04-13
fix: Better handling the import error hint

jdneo:
Since we will now display non Java projects (#744), the logic to check import error(#740) needs to be changed.

Now before we try to get the project nodes, we will first check if there is potential import errors. The extension will consider there is an import error if:

- No Java Project is imported into workspace except for default project
- There is at least one error marker in workspace.

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
@@ -296,25 +296,30 @@ public final class ProjectCommand {
 
     public static boolean checkImportStatus() {
         IProject[] projects = ProjectUtils.getAllProjects();
+        boolean hasJavaProjectImported = false;
+        boolean hasError = false;
         for (IProject project : projects) {
             if (ProjectsManager.DEFAULT_PROJECT_NAME.equals(project.getName())) {
                 continue;
             }
+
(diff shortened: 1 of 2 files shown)


jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
```diff
@@ -296,25 +296,30 @@ public static String getModuleName(IJavaProject project) {
...
-                    boolean hasError = markers.stream().anyMatch(m -> {
-                        return m.getAttribute(IMarker.SEVERITY, 0) == IMarker.SEVERITY_ERROR;
-                    });
-                    if (hasError) {
-                        return true;
-                    }
+                int maxProblemSeverity = project.findMaxProblemSeverity(null, true, IResource.DEPTH_ONE);
+                if (maxProblemSeverity == IMarker.SEVERITY_ERROR) {
+                    hasError = true;
+                    break;
```
CsCherrYY, line 314: Given a non-java project, could it have an error marker? If so, we will directly return `true` here. However, we should return `false` if there is another Java project exists.

jdneo: In line 322, it will only return true when no Java projects found in workspace.

But this reminds me that I should early return if a Java project is found in the for loop.

CsCherrYY (approved):

PR #749, opened 2023-04-13
feat: Introduce a setting to control whether non-Java resources should be filtered

jdneo:
See:

https://user-images.githubusercontent.com/6193897/231690524-4bfd3060-e039-43b7-8df3-a6e737b43499.mp4

The next step is add a button to quickly switch the behavior. It's not implemented yet due to an issue in VS Code: https://github.com/microsoft/vscode/issues/179842


Eskibear (dismissed):

jdneo:
People from VS Code says that it is by design that the property of an object setting will not have the setting change update event.

If we want to expose this setting to a command, we need to change it.

I'm thinking to change it to: java.project.explorer.showNonJavaResources. Will append a PR to address that.

jdneo:
New UX: Give up using two commands toggling each other. Now using a static command to control all the filters in case that in the future new filters will be added.


https://user-images.githubusercontent.com/6193897/231926944-794b2f75-2d40-4ae3-bb83-3cb83f64c3f5.mp4

src/views/dependencyExplorer.ts
```diff
@@ -90,6 +90,23 @@ export class DependencyExplorer implements Disposable {
...
                 this.reveal(uri, false /*force to reveal even the sync setting is turned off*/);
             }),
+            instrumentOperationAsVsCodeCommand(Commands.JAVA_PROJECT_EXPLORER_CONFIGURE_FILTERS, async () => {
+                const filters = await window.showQuickPick(
+                    [{
+                        label: "Non-Java resources",
+                        picked: Settings.nonJavaResourcesFiltered(),
+                    }],
+                    {
+                        placeHolder: "Select filters to apply to the Java Projects Explorer",
```
CsCherrYY, line 100: add `ignoreFocusOut`? since users may have chance to check the files in the explorer.

jdneo: Ok, added

CsCherrYY (approved):
LGTM

PR #801, opened 2023-11-28
feat: Improve the user experience of new Java file.

testforstephen:
The new file experience will ask user to select the file type first, then input the type name and generate the new file.

https://github.com/microsoft/vscode-java-dependency/assets/14052197/587148d1-c896-4c02-ada5-c4fd7063f3ca


jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java
```diff
@@ -173,6 +180,10 @@ public static PackageNode createNodeForProject(IJavaElement javaElement) {
                 projectNode.setMetaDataValue(UNMANAGED_FOLDER_INNER_PATH, proj.getLocationURI().toString());
             }
             projectNode.setMetaDataValue(NATURE_ID, natureIds);
+            String sourceVersion = javaElement.getJavaProject().getOption(JavaCore.COMPILER_SOURCE, true);
+            int jdkLevel = (int) (CompilerOptions.versionToJdkLevel(sourceVersion, true) >>> 16);
```
jdneo, line 184: Why cannot directly use the COMPILER_SOURCE value? To change `1.8`` to `8`?

jdneo:
What about adding a divider between package & file, folder?

image

image

package.json
```diff
@@ -895,6 +970,7 @@
   },
   "dependencies": {
     "await-lock": "^2.2.2",
+    "fmtr": "^1.1.4",
```
jdneo, line 973: Add its license to third party notice.

src/views/PrimaryTypeNode.ts
```diff
@@ -121,6 +121,11 @@ export class PrimaryTypeNode extends DataNode {
             contextValue += "+test";
         }
 
+        if (this._rootNode?.getParent() instanceof ProjectNode
+            && (this._rootNode.getParent() as ProjectNode).nodeData?.metaData?.MaxSourceVersion >= 14) {
```
jdneo, line 125: Given that record is a preview feature in Java 14 and become a standard feature starting from Java 16. I think we should compare with `16` here.

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java
testforstephen: yes, it's used to change it to an int value.

src/views/PrimaryTypeNode.ts
testforstephen: yes, makes sense. The latest Java extension doesn't support preview for the old versions like 14 and 15, it makes sense to enable it only on Java 16 and higher.

nickzhums: yes, makes sense

src/explorerCommands/new.ts
```diff
@@ -186,6 +339,62 @@ function canCreateClass(node: DataNode): boolean {
...
+    }
+
+    if (!sourcePaths || !sourcePaths.length) {
+        return "";
+    }
+
+    for (const sourcePath of sourcePaths) {
+        if (isPrefix(sourcePath, filePath)) {
+            const relative = path.relative(sourcePath, path.dirname(filePath));
+            return relative.replace(/[\/\\]/g, ".");
```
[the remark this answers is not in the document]
testforstephen: Great. Verified that `/` is not required to escape.

testforstephen:
> What about adding a divider between package & file, folder?

Here are new menus view.
image

image

jdneo (approved):

PR #820, opened 2024-02-22
Add 'New Java File' menu to file explorer & adjust the existing file explorer menu order

testforstephen:
- Add 'New Java File' menu to file explorer
- Put the existing menus such as 'Reveal in Java Project Explorer', 'New Java Project...' into the same category as other Java context menus.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -547,17 +547,22 @@
         {
           "command": "java.view.package.revealInProjectExplorer",
           "when": "resourceFilename =~ /(.*\\.gradle)|(.*\\.gradle\\.kts)|(pom\\.xml)$/ && java:serverMode == Standard",
-          "group": "navigation@100"
+          "group": "1_javaactions"
         },
         {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -791,6 +796,38 @@
           "group": "new2@40",
           "when": "view == javaProjectExplorer && (viewItem =~ /java:(file|folder|project)/ || viewItem =~ /java:(packageRoot)(?=.*?\\b\\+resource\\b)/)"
         }
+      ],
+      "javaProject.newJavaFile": [
+        {
+          "command": "java.view.package.newJavaClass",
+          "group": "new@10",
+          "when": "explorerResourceIsFolder"
```
jdneo, line 804: Any reason that we do not take file node into consideration?

testforstephen: this is same as the behavior of the standard VS Code "New File/Folder ..." menu, which only appears on the folder context.

jdneo: ok

Issue #857, opened 2024-10-03
[Improvement] - Change dependency jar naming to 'groupId:artifactId:version' for maven projects


mamilic:
As of now Referenced libraries are shown as list with jar names, and sometimes it is hard to guess what that is, for example we have **core-3.4.0.jar**, or **context-2.3.2.jar**, picture below. 

![image](https://github.com/user-attachments/assets/19edab42-467a-4912-b986-f0c1fe7d4c11)

Is it possible that to have something which intellij uses, to have groupId:artifactId:version?

![image](https://github.com/user-attachments/assets/81144613-2c0f-4a6b-933a-4c65cc114e78)

jdneo:
Sounds a valid request

mamilic:
@jdneo , is this perhaps good first issue?

jdneo:
@mamilic If we can easily get the group id, artifact id and version from the jar files, then this should be a good first issue.

I need to find some time to investigate it. It would be great if you are willing to have a try :)

mamilic:
@jdneo , I can try. But if you have any hints where to look, please let me know. Thanks!

jdneo:
I've a question actually. Does every maven jar contains this `pom.properties`?

Seems it's not in some cases:

![Image](https://github.com/user-attachments/assets/44560ed3-3a81-443d-ac8f-34bd777ca994)

mamilic:
It seems that you are correct, however as I observed we would always have MANIFEST.MF. Could it be possible that JDTLS somehow proivides more info about the jars/dependencies? Also, I am wondering how may I run the extension in debug mode, when I try to launch it from VSCode it starts but then throws 
`rejected promise not handled within 1 second: Error: No delegateCommandHandler for java.resolvePath
extensionHostProcess.js:162
stack trace: Error: No delegateCommandHandler for java.resolvePath
	at /home/mmilic/.vscode/extensions/redhat.java-1.35.1-linux-x64/dist/extension.js:2:1090351
	at re (/home/mmilic/.vscode/extensions/redhat.java-1.35.1-linux-x64/dist/extension.js:2:1090645)
	at /home/mmilic/.vscode/extensions/redhat.java-1.35.1-linux-x64/dist/extension.js:2:1085429
	at Immediate.<anonymous> (/home/mmilic/.vscode/extensions/redhat.java-1.35.1-linux-x64/dist/extension.js:2:1085449)
	at process.processImmediate (node:internal/timers:483:21)
[...]

PR #859, opened 2024-10-16
ux - display maven and gradle dependencies with pattern 'groupId:artifactId:version '

mamilic:
This MR referes to the issue, https://github.com/microsoft/vscode-java-dependency/issues/857

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,31 @@
+# How to Contribute
+
+We greatly appreciate contributions to the vscode-java-dependency project. Your efforts help us maintain and improve this extension. To ensure a smooth contribution process, please follow these guidelines.
+
+## Setup Guide
+
+To set up the vscode-java-dependency project, follow these steps:
+
+1. **Build the Server JAR**:
+   - The server JAR (Java application) is located in the [jdtls.ext](./jdtls.ext) directory.
+   - Run the following command to build the server:
+     ```shell
+     npm run build-server
+     ```
+
+2. **Install Dependencies**:
+   - Execute the following command to install the necessary dependencies:
+     ```shell
+     npm install
+     ```
+
+3. **Run the Extension**:
+   - Open the "Run and Debug" view in Visual Studio Code.
+   - Run the "Run Extension" task.
+
+4. **Attach to Plugin**:
+   - Open the "Run and Debug" view in Visual Studio Code.
+   - Run the "Attach to Plugin" task.
+   - Note: This task is required only if you want to debug Java code [jdtls.ext](./jdtls.ext). It requires the [vscode-pde](https://marketplace.visualstudio.com/items?itemName=yaozheng.vscode-pde) extension to be installed.
+
+Thank you for your contributions and support!
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,12 @@
 
 We greatly appreciate contributions to the vscode-java-dependency project. Your efforts help us maintain and improve this extension. To ensure a smooth contribution process, please follow these guidelines.
 
-## Setup Guide
+## Prerequisites
+- [JDK](https://www.oracle.com/java/technologies/downloads/?er=221886)
+- [Node.JS](https://nodejs.org/en/)
+- [VSCode](https://code.visualstudio.com/)
+
+## Build and Run
 
 To set up the vscode-java-dependency project, follow these steps:
 
@@ -19,11 +24,12 @@ To set up the vscode-java-dependency project, follow these steps:
      npm install
      ```
 
-3. **Run the Extension**:
+3. **Run/Debug the Extension**:
    - Open the "Run and Debug" view in Visual Studio Code.
    - Run the "Run Extension" task.
 
-4. **Attach to Plugin**:
+4. **Attach to Plugin[Debug Java]**:
+   - Prerequisite: Ensure that the extension is activated, meaning the Java process is already launched. This is required for the task to run properly.
    - Open the "Run and Debug" view in Visual Studio Code.
    - Run the "Attach to Plugin" task.
    - Note: This task is required only if you want to debug Java code [jdtls.ext](./jdtls.ext). It requires the [vscode-pde](https://marketplace.visualstudio.com/items?itemName=yaozheng.vscode-pde) extension to be installed.
--- a/src/views/PrimaryTypeNode.ts
+++ b/src/views/PrimaryTypeNode.ts
@@ -34,6 +34,10 @@ export class PrimaryTypeNode extends DataNode {
         return "";
     }
 
+    public getLabel(): string {
+        return this._nodeData.displayName ?? this._nodeData.name;
+    }
+
     protected async loadData(): Promise<SymbolInformation[] | DocumentSymbol[] | undefined> {
         if (!this.hasChildren() || !this.nodeData.uri) {
             return undefined;
--- a/src/views/containerNode.ts
+++ b/src/views/containerNode.ts
@@ -43,8 +43,8 @@ export class ContainerNode extends DataNode {
         return this._containerType;
     }
 
-    public getLabel(): string {
-        return this._nodeData.displayName ?? this._nodeData.name;
+    public isMavenType(): boolean {
+        return this._containerType == ContainerType.Maven;
     }
 
     protected async loadData(): Promise<INodeData[]> {
--- a/src/views/containerNode.ts
+++ b/src/views/containerNode.ts
@@ -44,11 +44,7 @@ export class ContainerNode extends DataNode {
     }
 
     public getLabel(): string {
-        if (this._nodeData.metaData?.['maven.groupId']) {
-            return `${this._nodeData.metaData?.['maven.groupId']}:${this._nodeData.metaData?.['maven.artifactId']}:${this._nodeData.metaData?.['maven.version']}`;
-        } else {
-            return this._nodeData.displayName ?? this._nodeData.name;
-        }
+        return this._nodeData.displayName ?? this._nodeData.name;
     }
 
     public isMavenType(): boolean {
--- a/src/views/containerNode.ts
+++ b/src/views/containerNode.ts
@@ -15,23 +15,48 @@ export class ContainerNode extends DataNode {
         super(nodeData, parent);
     }
 
+    private _containerType: ContainerType;
+
     public get projectBasePath() {
         return this._project.uri && Uri.parse(this._project.uri).fsPath;
     }
 
-    public getContainerType(): string {
+    public getContainerType(): ContainerType {
+        if (this._containerType) {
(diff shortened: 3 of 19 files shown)


mamilic:
Hi @jdneo , this is my view of the implementation for https://github.com/microsoft/vscode-java-dependency/issues/857, though I am not fully satisfied with changes, as I had to introduce getLabel() method to ExplorerNode. 

Could you please take a look and let me know what I might have done better? Also, is there a requirement to test his functionality?

src/views/containerNode.ts
```diff
@@ -15,23 +15,48 @@ export class ContainerNode extends DataNode {
...
+            this._containerType = ContainerType.Unknown;
         }
-        return ContainerType.Unknown;
+
+        return this._containerType;
+    }
+
+    public getLabel(): string {
+        if (this._nodeData.metaData?.['maven.groupId']) {
+            return `${this._nodeData.metaData?.['maven.groupId']}:${this._nodeData.metaData?.['maven.artifactId']}:${this._nodeData.metaData?.['maven.version']}`;
```
jdneo, line 48: Just an ask:

Those metadata only exist for maven project, since they are provided by the maven project importer. Which means the dependencies of gradle project could not have the new display style. Is it as expected for you?

mamilic: That is correct, however I thought that those are the same for Gradle, but it seems they are not. It only works for Maven, as when gradle project is loaded, the jdtls is not recognizing dependencies as Gradle, instead it is.
image

Do you know maybe where can I look to resolve this?

jdneo: Because in JDT.LS, maven project importer and gradle project importer are two different components. So far only maven project importer leaves those metadata while the gradle project importer does not.

I've no idea how we can achieve further given the current situation. Let me think a little bit.

mamilic: Could you point me in direction where I can look at that implementation for maven and gradle importer?

jdneo: That will be a little bit complex.

This is maven importer: https://github.com/eclipse-m2e/m2e-core

For gradle importer we have two options:

- https://github.com/microsoft/build-server-for-gradle & https://github.com/microsoft/vscode-gradle/tree/develop/extension/jdtls.ext/com.microsoft.gradle.bs.importer (recommended one right now)
- https://github.com/eclipse/buildship

They are separate independent projects. I think you can check if we can do similar thing as maven at the gradle importer side. Check if we can add those metadata into classpath entries [here](https://github.com/microsoft/vscode-gradle/blob/9542e919a4b29c6ffff683c3991e3e498fdfd34a/extension/jdtls.ext/com.microsoft.gradle.bs.importer/src/com/microsoft/gradle/bs/importer/GradleBuildServerBuildSupport.java#L643)

jdneo: If you are willing to address the gradle side issue, I think we can put it into another PR/issue. And focus the maven side in this one.

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+     npm install
+     ```
+
+3. **Run the Extension**:
+   - Open the "Run and Debug" view in Visual Studio Code.
+   - Run the "Run Extension" task.
+
+4. **Attach to Plugin**:
+   - Open the "Run and Debug" view in Visual Studio Code.
+   - Run the "Attach to Plugin" task.
```
jdneo, line 28: A prerequisite of this launch configuration is that the extension is activated, in other words, the java process is launched. Let's add this information as well in the document.

src/views/containerNode.ts
mamilic: Would it be feasible to have separate display of jar's names for Maven and Gradle? I'll try to address the Gradle side as well in timely manner.

jdneo: This is exactly what I mean

src/views/packageRootNode.ts
```diff
@@ -20,6 +20,16 @@ export class PackageRootNode extends DataNode {
         super(nodeData, parent);
     }
 
+    public getLabel(): string {
+        if (this.getParent() instanceof ContainerNode) {
+            const parent = this.getParent() as ContainerNode;
+            if (parent.getContainerType() == ContainerType.Maven) {
```
mamilic, line 26: @jdneo , I've noticed that PackageRootNode is actually type of Java Package that Label needs to change, would this be feasable? I can check if it is a Maven type by looking at parent if it is ContainerNode, or other approach would be just to check for metaData.['maven.groupId']. What is more favourable approach?

CONTRIBUTING.md
mamilic: Updated.

src/views/packageRootNode.ts
jdneo: So far just check `just to check for metaData.['maven.groupId']` looks good to me.

mamilic: Done.

CONTRIBUTING.md
```diff
@@ -0,0 +1,37 @@
+# How to Contribute
```
jdneo, line 1: Thank you for drafting the content of this document 👍

src/views/dependencyDataProvider.ts
```diff
@@ -124,6 +124,14 @@ export class DependencyDataProvider implements TreeDataProvider<ExplorerNode> {
         const children = (!this._rootItems || !element) ?
             await this.getRootNodes() : await element.getChildren();
 
+        if (children && element instanceof ContainerNode) {
+            if (element.isMavenType()) {
```
jdneo, line 128: I think it makes sense to always sort alphabetically. Then there is no need check the type of the container. And we can remove `isMavenType()` & `isGradleType()`

mamilic: Done.

mamilic: @jdneo , how would you like me to name metaData for gradle, do we want to name them same as the maven one? maven.groupId, maven.artifactId, maven.version?

jdneo: Ah that's a good question.

I think still calling it `maven.xxx` does not make sense anymore. Since they are not persisted by maven importer. Meanwhile, a dependency may not come from maven registry.

what about just calling it `groupId `, `artifactId ` and `version`?

BTW, since it's specific with gradle importer, we can create an issue and discuss those implementation details there.

src/views/containerNode.ts
```diff
@@ -15,23 +15,36 @@ export class ContainerNode extends DataNode {
         super(nodeData, parent);
     }
 
+    private _containerType: ContainerType;
```
jdneo, line 18: It's not used for now, how about removing it?

src/views/PrimaryTypeNode.ts
```diff
@@ -34,6 +34,10 @@ export class PrimaryTypeNode extends DataNode {
         return "";
     }
 
+    public getLabel(): string {
```
jdneo, line 37: Instead of introducing `getLabel()` into all different types. What about directly set the displayName at java side?

src/views/containerNode.ts
mamilic: I was thinking to leave it as it will "cache" the outcome of the **getContainerType()** method. Does it make sence then to still be present?

jdneo: OK I see. Yeah it makes sense!

src/views/dependencyDataProvider.ts
```diff
@@ -124,6 +124,14 @@ export class DependencyDataProvider implements TreeDataProvider<ExplorerNode> {
         const children = (!this._rootItems || !element) ?
             await this.getRootNodes() : await element.getChildren();
 
+            if (children && element instanceof ContainerNode) {
```
mamilic, line 127: @jdneo We might want to leave this check, as without it, it will sort everything always.
image

src/views/dependencyDataProvider.ts
```diff
@@ -124,6 +124,14 @@ export class DependencyDataProvider implements TreeDataProvider<ExplorerNode> {
...
+                if (element.isMavenType()) {
+                    children.sort((a, b) => {
+                        return a.getDisplayName().localeCompare(b.getDisplayName());
```
mamilic, line 130: @jdneo, we will still need displayName or label method to be available here, as we need to sort tha array.

src/views/dependencyDataProvider.ts
jdneo: ok, makes sense.

src/views/dependencyDataProvider.ts
jdneo: Makes sense. I now think that sorting the nodes at TS side is valid. Because in the future may be we will introduce some kind of settings controlling the sort approach in the tree. And the TS (client) side should change the sort behavior based on the setting (which is also a client side stuff.)

src/views/containerNode.ts
```diff
@@ -15,23 +15,36 @@ export class ContainerNode extends DataNode {
...
+        } else {
+            this._containerType = ContainerType.Unknown;
         }
-        return ContainerType.Unknown;
+
+        return this._containerType;
+    }
+
+    public isMavenType(): boolean {
+        return this._containerType == ContainerType.Maven;
```
jdneo, line 47: ERROR: (triple-equals) D:/a/vscode-java-dependency/vscode-java-dependency/src/views/containerNode.ts[47, 36]: == should be ===

jdneo:
Overall looks good to me. Are you willing to add a test case in the `maven-suite/projectView.test.ts`? I can also handle that you have no time to do that.

mamilic:
> Overall looks good to me. Are you willing to add a test case in the `maven-suite/projectView.test.ts`? I can also handle that you have no time to do that.

I'm definitely willing to add a test case. It would be a great learning opportunity for me, though it might take some time. That said, it depends on how quickly we’re looking to merge this. Let me know what works best!

jdneo:
I'm not hurry, just take your time 😉

mamilic:
@jdneo, 2 tests are failing, though I am not sure if that is intended, as the latest build from main branch reflect the same result, https://github.com/microsoft/vscode-java-dependency/actions/runs/10659297304/job/29541618151

jdneo:
It's unlikely the failures are caused by this PR. Please give me some time to do some investigation (I was occupied by some other stuff recently).

If I cannot get some finding I'll merge this PR first.

jdneo:
I tested on my mac but I cannot reproduce it.

jdneo (approved):
LGTM

jdneo:
@mamilic Thanks for your contribution!

I appended a commit to pass the test. I guess it's related with some timing issue. Since gradle project needs to download gradle distribution during import in CI jobs. It will take longer time.

mamilic:
@jdneo, thank you for your patience and guidence on this issue, and do apologies for taking so long. Next step is to enable this for gradle.

PR #901, opened 2025-09-03
Remind users to upgrade old (<21) Java and EOL Spring Boot/Framework versions

FluoriteCafe-work:
Components for Java engine/dependency upgrade reminder.


src/views/workspaceNode.ts
```diff
@@ -18,7 +19,9 @@ export class WorkspaceNode extends DataNode {
         if (!this.nodeData.uri) {
             return undefined;
         }
-        return Jdtls.getProjects(this.nodeData.uri);
+        const result = await Jdtls.getProjects(this.nodeData.uri);
+        PromotionProvider.getInstance().checkJavaVersion(result);
+        return result;
```
wangmingliang-ms: 1. what if this method is called >1 times?
2. The loadData function should only handle data retrieval. Performing a Java version validation at the same time as data fetching seems quite unreasonable. This is clearly a "side effect" and should be avoided.

FluoriteCafe-work: > what if this method is called >1 times?

Controls are [implemented in a downstream function](https://github.com/microsoft/vscode-java-dependency/pull/901/files#diff-48d127a7e1dcd41c82fc2f22babbec1434bc304ca5b1babfa5ed7327bad3a1edR32-R35) to prevent repeated notifications.

> The loadData function should only handle data retrieval.

I'll seek to move the added logic out of `loadData()`.

FluoriteCafe-work: I've updated the PR to check Java version on creation of `ProjectNode` (`new ProjectNode(...)`) - Java version is related to this type of node. Outside the scope the Java version data is hard to be read (it's saved in a `protected` property) without touching the property structure of classes.

FluoriteCafe-work:
Current component dependency tree:

```mermaid
flowchart TD
    DiagnosticsManager
    NotificationManager
    IssueManager
    MetadataManager
    UpgradeManager
    UpgradeCodeActionProvider
    UpgradeManager --> MetadataManager
    UpgradeManager --> IssueManager
    UpgradeCodeActionProvider --> MetadataManager
    UpgradeCodeActionProvider --> DiagnosticsManager
    IssueManager --> DiagnosticsManager
    IssueManager --> NotificationManager
```

package.nls.json
akaroml, line 1: Please only make necessary changes to reduce the size of of the PR. In this file, the only change is the inserted line of:
```
  "contributes.commands.java.view.triggerJavaUpgradeTool": "Upgrade dependencies",
```

src/constants.ts
```diff
@@ -35,6 +35,13 @@ export namespace ExtensionName {
     export const JAVA_LANGUAGE_SUPPORT: string = "redhat.java";
 }
 
+export namespace Upgrade {
+    export const DIAGNOSTICS_GROUP_ID_FOR_JAVA_ENGINE = "java-engine";
+    export const DIAGNOSTICS_NAME_FOR_JAVA_ENGINE = "Java Engine";
+    export const PROMOTION_DIAGNOSTIC_SOURCE = "Java";
+    export const EARLIEST_JAVA_VERSION_NOT_TO_PROMPT = 21;
```
akaroml: LATEST_JAVA_LTS_VERSION

src/constants.ts
```diff
@@ -35,6 +35,13 @@ export namespace ExtensionName {
     export const JAVA_LANGUAGE_SUPPORT: string = "redhat.java";
 }
 
+export namespace Upgrade {
+    export const DIAGNOSTICS_GROUP_ID_FOR_JAVA_ENGINE = "java-engine";
+    export const DIAGNOSTICS_NAME_FOR_JAVA_ENGINE = "Java Engine";
```
akaroml: Use "java-upgrade-assistant" and "Java Upgrade Assistant" instead.

src/constants.ts
```diff
@@ -35,6 +35,13 @@ export namespace ExtensionName {
...
+    export const PROMOTION_DIAGNOSTIC_SOURCE = "Java";
```
akaroml: "Java Upgrade Assistant"

src/extension.ts
```diff
@@ -2,8 +2,10 @@
 // Licensed under the MIT license.
 
 import * as path from "path";
-import { commands, Diagnostic, Extension, ExtensionContext, extensions, languages,
-    Range, tasks, TextDocument, TextEditor, Uri, window, workspace } from "vscode";
+import {
```
akaroml: Please avoid unnecessary changes.

src/syncHandler.ts
```diff
@@ -26,7 +27,7 @@ class SyncHandler implements Disposable {
         if (autoRefresh) {
             instrumentOperation(ENABLE_AUTO_REFRESH, () => this.enableAutoRefresh())();
         } else {
-            instrumentOperation(DISABLE_AUTO_REFRESH, () => {})();
+            instrumentOperation(DISABLE_AUTO_REFRESH, () => { })();
```
akaroml: Avoid unnecessary changes.

src/syncHandler.ts
```diff
@@ -122,7 +124,7 @@ class SyncHandler implements Disposable {
         } else {
             // in flat view
             if (path.extname(uri.fsPath) === ".java" && node.uri &&
-                    Uri.parse(node.uri).fsPath === path.dirname(uri.fsPath)) {
```
akaroml: Avoid unnecessary changes.

akaroml (commented):
Please detect whether the App Mod ext is active. If it's inactive, no hint should be shown to the users.

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,150 @@
...
+        });
+    }
+
+    public scan() {
+        workspace.workspaceFolders?.forEach((folder) =>
+            this.checkUpgradableComponents(folder)
+        );
+    }
+
+    private async checkUpgradableComponents(folder: WorkspaceFolder) {
```
akaroml: Since we are already in Project Explorer, we already know all the dependencies and the version of java runtimes. Can we work with the project metadata directly? wihtout parsing the POM/Gradle files manually? CC @jdneo

akaroml (commented):
In general, I wasn't expecting so many code changes. The fat implementation of manually parsing everything is the root cause. As commented, please consider work with the project metadata directly. In the metadata, the runtime version and dependency versions should all be available.

akaroml (changes requested):
In general, I wasn't expecting so many code changes. The fat implementation of manually parsing everything is the root cause. As commented, please consider work with the project metadata directly. In the metadata, the runtime version and dependency versions should all be available.

FluoriteCafe-work:
@akaroml Thanks for the review! 

> As commented, please consider work with the project metadata directly. In the metadata, the runtime version and dependency versions should all be available.

The code is now working with `Jdtls` to retrieve all metadata. The extension has already gone through the process once, but it's in a lazy-loading way - only retrieving needed parts (projects and dependencies etc.) when user expands the nodes in tree view. As a result, the PR need to actively call `Jdtls` by itself to gain a full picture.

package.json
```diff
@@ -568,6 +583,11 @@
           "when": "explorerResourceIsFolder",
           "group": "1_javaactions@30"
         },
+        {
+          "command": "java.view.modernizeJavaProject",
+          "when": "explorerResourceIsFolder && isModernizationExtensionInstalled",
```
wangmingliang-ms: && is java project.

package.json
```diff
@@ -323,6 +333,11 @@
           "description": "%configuration.java.dependency.packagePresentation%",
           "default": "flat"
         },
+        "java.dependency.showUpgradeReminder": {
```
wangmingliang-ms: how about "java.dependency.enableDependencyDiagnosing"?
note that this will also be rendered as UI text in user settings.

package.nls.json
```diff
@@ -24,6 +24,8 @@
   "contributes.commands.java.view.package.copyRelativeFilePath": "Copy Relative Path",
   "contributes.commands.java.view.package.new": "New...",
   "contributes.commands.java.view.package.newJavaClass": "Class...",
+  "contributes.commands.java.view.triggerJavaUpgradeTool": "Upgrade dependencies",
```
wangmingliang-ms: "Upgrade runtime & frameworks"? to be consistent with the button title.

src/commands.ts
```diff
@@ -42,6 +42,10 @@ export namespace Commands {
 
     export const VIEW_PACKAGE_NEW_JAVA_CLASS = "java.view.package.newJavaClass";
 
+    export const VIEW_TRIGGER_JAVA_UPGRADE_TOOL = "java.view.triggerJavaUpgradeTool";
```
wangmingliang-ms: I suspect that, according to convention, the prefix of a command should be related to the location where the context menu is displayed. Therefore, the prefixes of these two commands are likely incorrect.

src/constants.ts
```diff
@@ -33,6 +33,15 @@ export namespace Explorer {
...
 export namespace ExtensionName {
     export const JAVA_LANGUAGE_SUPPORT: string = "redhat.java";
+    export const APP_MODERNIZATION_FOR_JAVA = "vscjava.migrate-java-to-azure";
+    export const APP_MODERNIZATION_UPGRADE_FOR_JAVA = "vscjava.vscode-java-upgrade";
+}
+
+export namespace Upgrade {
+    export const DIAGNOSTICS_GROUP_ID_FOR_JAVA_RUNTIME = "java-upgrade-assistant";
+    export const DIAGNOSTICS_NAME_FOR_JAVA_RUNTIME = "Java Upgrade Assistant";
+    export const LATEST_JAVA_LTS_VESRION = 21;
```
wangmingliang-ms: why is only Java here, where is SpringBoot...? is this used outside metadataManager?

package.json
```diff
@@ -288,6 +288,16 @@
         "command": "java.view.package.renameFile",
         "title": "%contributes.commands.java.view.package.renameFile%",
         "category": "Java"
+      },
+      {
+        "command": "java.view.triggerJavaUpgradeTool",
```
wangmingliang-ms: how about "xxx.upgradeRuntimeAndFrameworks" or "xxx.upgradeWithCopilot"

src/settings.ts
```diff
@@ -108,6 +108,10 @@ export class Settings {
         return workspace.getConfiguration("java.dependency").get("refreshDelay", 2000);
     }
 
+    public static getShowUpgradeReminder() {
```
wangmingliang-ms: `isDependencyDiagnosingEnabled()`?

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java
```diff
@@ -181,6 +183,10 @@ public static PackageNode createNodeForProject(IJavaElement javaElement) {
             int jdkLevel = (int) (CompilerOptions.versionToJdkLevel(sourceVersion, true) >>> 16);
             int majorVersion = Math.max(0, jdkLevel - ClassFileConstants.MAJOR_VERSION_0);
             projectNode.setMetaDataValue(MAX_SOURCE_VERSION, majorVersion);
+            IFile existingPom = proj.getFile("pom.xml");
+            if (existingPom.exists()) {
+                projectNode.setMetaDataValue(POM_PATH, existingPom.getLocation().toOSString());
+            }
```
wangmingliang-ms: is this still needed? we don't need to show diagnostics according to the adjusted scope.

src/upgrade/display/notificationManager.ts
```diff
@@ -0,0 +1,86 @@
...
+        return this.context?.globalState.update(SESSION_COUNT_KEY, num);
+    }
+
+    private isCandidate() {
+        return this.context?.globalState.get<boolean>(IS_CANDIDATE_KEY, true);
+    }
+
+    private setCandidate(isCandidate: boolean) {
+        this.context?.globalState.update(IS_CANDIDATE_KEY, isCandidate);
+    }
```
wangmingliang-ms: not used?

src/constants.ts
```diff
@@ -33,6 +33,15 @@ export namespace Explorer {
...
+    export const SESSION_COUNT_BEFORE_NOTIFICATION_RESHOW = 3;
```
wangmingliang-ms: is this also used outside the notificationManager?

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: I realized it should actually be changed to read from/write to the setting item `java.dependency.enableDependencyDiagnosing`. Will change later.

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java
FluoriteCafe-work: Nope for now. I'll remove parts related to POM path parsing.

src/commands.ts
FluoriteCafe-work: I'm also thinking whether the command would fit in any of the current namespace (`java.view` or `java.project`). Could you give some suggestions? Maybe it can use a new namespace like `java.upgrade.triggerTool` or so on.

package.nls.json
FluoriteCafe-work: Seems better - BTW do you think we should not expose the command to Command Palette (as it only uses a generic prompt `DEFAULT_UPGRADE_PROMPT`)?

package.json
FluoriteCafe-work: Well, I've been trying to avoid mentioning the word `dependency` in strings as we also check Java runtime issues... not sure whether it may be somehow confusing.

src/upgrade/dependency.data.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Upgrade } from "../constants";
+import { UpgradeReason, type DependencyCheckMetadata } from "./type";
+
+export const DEPENDENCY_JAVA_RUNTIME = {
+    "name": "Java Runtime",
+    "reason": UpgradeReason.JRE_TOO_OLD,
+    "supportedVersion": `>=${Upgrade.LATEST_JAVA_LTS_VESRION}`,
+} as const;
+
+const DEPENDENCIES_TO_SCAN: DependencyCheckMetadata = {
```
FluoriteCafe-work: TODO: Add `suggestedVersion` here

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,192 @@
...
+import { Commands } from "../commands";
+import metadataManager from "./metadataManager";
+import { buildPackageId } from "./utility";
+import notificationManager from "./display/notificationManager";
+import { Settings } from "../settings";
+import { DEPENDENCY_JAVA_RUNTIME } from "./dependency.data";
+
+const DEFAULT_UPGRADE_PROMPT = "Upgrade Java project dependency to latest version.";
+
+function getJavaIssues(data: INodeData): UpgradeIssue[] {
```
FluoriteCafe-work: TODO: Pick issue detection logic out to a separate file (e.g. `assessmentManager`)

src/upgrade/display/notificationManager.ts
```diff
@@ -0,0 +1,86 @@
...
+
+class NotificationManager {
+    private hasShown = false;
+    private context?: ExtensionContext;
+
+    initialize(context: ExtensionContext) {
+        this.context = context;
+    }
+
+    async triggerNotification(issue: UpgradeIssue) {
```
FluoriteCafe-work: TODO: make it an interface

src/constants.ts
```diff
@@ -33,6 +33,14 @@ export namespace Explorer {
 
 export namespace ExtensionName {
     export const JAVA_LANGUAGE_SUPPORT: string = "redhat.java";
+    export const APP_MODERNIZATION_FOR_JAVA = "vscjava.migrate-java-to-azure";
+    export const APP_MODERNIZATION_UPGRADE_FOR_JAVA = "vscjava.vscode-java-upgrade";
+}
+
+export namespace Upgrade {
+    export const PACKAGE_ID_FOR_JAVA_RUNTIME = "java-runtime:*";
```
wangmingliang-ms: `java:*`

src/commands.ts
FluoriteCafe-work: TODO: hide the command from Palette

src/upgrade/upgradeManager.ts
wangmingliang-ms: telemeter detected upgrade issues.

dependency name
current version

src/upgrade/display/notificationManager.ts
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+
+        const prompt = buildFixPrompt(issue);
+        const notificationMessage = buildNotificationMessage(issue);
+        const selection = await window.showInformationMessage(
+            notificationMessage,
+            BUTTON_TEXT_UPGRADE,
+            BUTTON_TEXT_NOT_NOW,
+            BUTTON_TEXT_DONT_SHOW_AGAIN);
```
wangmingliang-ms: telemeter user actions.

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: The button should change to workspace-level setting rather than user-level setting.

package.json
FluoriteCafe-work: ~~The extension only activates in Java(-like) workspaces, so I think we don't need to do this check here manually.~~

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: Fixed in e25d0b6.

src/upgrade/upgradeManager.ts
FluoriteCafe-work: Fixed in e25d0b6 & 46f78a8.

src/constants.ts
FluoriteCafe-work: Fixed in 76a8261a7094fa9366d86eee76d1a62cc29486ad.

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: 790a5eaf6fcbcde982dadfd066c2c793b90a43e2

src/upgrade/upgradeManager.ts
FluoriteCafe-work: fbe2d5fb760d3aaf65abceb8f8b6a911cb0028a0

src/upgrade/dependency.data.ts
FluoriteCafe-work: 56084e63c1fbcdc71c7fa1def1051c2723c2d18c

src/constants.ts
FluoriteCafe-work: Moved most strings in their own files - 56084e63c1fbcdc71c7fa1def1051c2723c2d18c.

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: Fixed by 1f0a36f653297e54caada1d3c5cc2f06e5808758

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/model/PackageNode.java
FluoriteCafe-work: Removed in 2789537642e07fa966b7204169b6fa0ae0713ffc

src/settings.ts
FluoriteCafe-work: Changed to `Settings.getEnableDependencyDiagnostics()` - cabb4698557ece2da9ea143cce6a23ee2ce6f969

package.json
FluoriteCafe-work: Fixed in 81b9905815a11ec6389f7fa5451057c15439095f & 28d2ee469832149d76a7c8d1e9e029110873d15e

src/constants.ts
FluoriteCafe-work: Java related metadata are now moved into `dependency.data.ts`.

src/commands.ts
FluoriteCafe-work: Commands are hidden: 4bdc6d617e03f9e9a310276174a3903a32b073be, 81b9905815a11ec6389f7fa5451057c15439095f

package.nls.json
FluoriteCafe-work: String updated; command hidden via 81b9905815a11ec6389f7fa5451057c15439095f.

package.json
FluoriteCafe-work: Renamed to `java.dependency.enableDependencyDiagnostics` - cabb4698557ece2da9ea143cce6a23ee2ce6f969

src/constants.ts
FluoriteCafe-work: Done in 56084e63c1fbcdc71c7fa1def1051c2723c2d18c.

package.nls.json
FluoriteCafe-work: Formatting are restored in c2c4cf65d15547420af89c904aa66bf86644b85a.

src/constants.ts
FluoriteCafe-work: Changed to `java:*` to align with the java-upgrade extension.

src/extension.ts
FluoriteCafe-work: Fixed in c3ee610082d349c9a68d820b84c9a8e00c53a6c3.

src/upgrade/display/notificationManager.ts
```diff
@@ -0,0 +1,98 @@
...
+    private hasShown = false;
+    private context?: ExtensionContext;
+
+    initialize(context: ExtensionContext) {
+        this.context = context;
+    }
+
+    async render(issues: UpgradeIssue[]) {
+        return (instrumentOperation(
+            "notificationManager.render",
```
wangmingliang-ms: how about `java.dependency.showUpgradeNotification`?

src/upgrade/display/notificationManager.ts
```diff
@@ -0,0 +1,98 @@
...
+                        commands.executeCommand(Commands.JAVA_UPGRADE_WITH_COPILOT, prompt);
+                        sendInfo(operationId, {
+                            __event_name__: "javaDependency.notificationManager.runUpgrade",
+                        });
+                        break;
+                    }
+                    case BUTTON_TEXT_NOT_NOW: {
+                        this.setSessionCount(-1 * SESSION_COUNT_BEFORE_NOTIFICATION_RESHOW);
+                        sendInfo(operationId, {
+                            __event_name__: "javaDependency.notificationManager.notNow",
```
wangmingliang-ms: `java.dependency.upgradeNotification.notNow`?
to be distinguished from other notification actions in this extension.

FluoriteCafe-work: Done in 03c85fbfa68bd1def3d40cbb24a94ea877a9ff30.

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: Done in 03c85fbfa68bd1def3d40cbb24a94ea877a9ff30.

src/upgrade/type.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export type UpgradeIssue = {
+    packageId: string;
+    packageDisplayName?: string;
+    reason: UpgradeReason;
+    currentVersion: string;
+    suggestedVersion?: string;
+};
+
+export interface DisplayInterface {
```
wangmingliang-ms: `UpgradeIssueRenderer`?
no need to add `Interface` in the name.

src/upgrade/upgradeManager.ts
```diff
@@ -63,26 +63,33 @@ class UpgradeManager {
...
-            const issues = await assessmentManager.getProjectIssues(projectNode);
-            projectIssues.push(...issues);
-        }));
+                const projectIssues: UpgradeIssue[] = [];
+                const uri = folder.uri.toString();
+                const projects = await Jdtls.getProjects(uri);
+                await Promise.allSettled(projects.map(async (projectNode) => {
+                    const issues = await assessmentManager.getProjectIssues(projectNode);
+                    projectIssues.push(...issues);
+                    sendInfo(operationId, {
```
Flanker32: Seems will send multi info events in single checkUpgradableComponents? Shall we move it to L89?

src/upgrade/assessmentManager.ts
```diff
@@ -0,0 +1,104 @@
...
+            currentVersion: String(javaVersion),
+            reason,
+            suggestedVersion,
+        }];
+    }
+
+    return [];
+}
+
+function getUpgrade(versionString: string, supportedVersionDefinition: DependencyCheckItem): Omit<UpgradeIssue, "packageId"> | null {
```
wangmingliang-ms: `getDependencyIssue`?

src/upgrade/display/notificationManager.ts
```diff
@@ -0,0 +1,98 @@
...
+                    notificationMessage,
+                    BUTTON_TEXT_UPGRADE,
+                    BUTTON_TEXT_NOT_NOW,
+                    BUTTON_TEXT_DONT_SHOW_AGAIN);
+
+                switch (selection) {
+                    case BUTTON_TEXT_UPGRADE: {
+                        commands.executeCommand(Commands.JAVA_UPGRADE_WITH_COPILOT, prompt);
+                        sendInfo(operationId, {
+                            __event_name__: "java.dependency.upgradeNotification.runUpgrade",
```
Flanker32: no need to use name like '__event_name__', please refer https://github.com/microsoft/vscode-java-dependency/blob/852589ac0c3a9454829c23c780e1f843aa049498/src/build.ts#L49-L52

src/upgrade/upgradeManager.ts
FluoriteCafe-work: This will send one info event per project (as it seems that `Jdtls` might retrieve multiple projects). I think we may not want to merge them, since merging the dependency graph of multiple projects doesn't help with telemetry analysis.

src/upgrade/assessmentManager.ts
```diff
@@ -0,0 +1,104 @@
...
+                    projectUri: projectNode.uri,
+                    path: packageContainer.path,
+                });
+                packages.forEach(
+                    (pkg) => {
+                        issues.push(...getDependencyIssues(pkg))
+                    }
+                );
+            })
+    );
```
wangmingliang-ms: `getPackagesDependencyIssues`

src/upgrade/assessmentManager.ts
```diff
@@ -0,0 +1,104 @@
...
+                }
+            }
+        }
+
+    }
+
+    return null;
+}
+
+function getDependencyIssues(data: INodeData): UpgradeIssue[] {
```
wangmingliang-ms: `getPackageDependencyIssues`

src/upgrade/dependency.data.ts
wangmingliang-ms, line 1: please rename to 'dependency.metadata.ts' to be consistent with 'metadataManager'

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: Fixed in 12f8a9bfcfc31feaa3d599ea253bb6f55e7c08bb.

src/upgrade/assessmentManager.ts
FluoriteCafe-work: Fixed in 4bec02348610b94d1ff449caade464d672fb19a5.

src/upgrade/dependency.data.ts
FluoriteCafe-work: Fixed in f7f2d29fec298f4f270eba26745ab1d1affc317a.

src/upgrade/type.ts
FluoriteCafe-work: Fixed in 6af31226e6d042e514cdd068cc6920557c04c43e.

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+import notificationManager from "./display/notificationManager";
+import { Settings } from "../settings";
+import assessmentManager from "./assessmentManager";
+
+const DEFAULT_UPGRADE_PROMPT = "Upgrade Java project dependency to latest version.";
+
+
+function shouldCheckUpgrade() {
+    return Settings.getEnableDependencyDiagnostics()
+        && !!extensions.getExtension(ExtensionName.APP_MODERNIZATION_UPGRADE_FOR_JAVA);
```
wangmingliang-ms, line 20: should also check if enabled. please refer to the code block below.

```
    const appModExtensionId = 'vscjava.migrate-java-to-azure';
    const appMod = vscode.extensions.getExtension(appModExtensionId);

    // Check if extension is installed and enabled
    const configuration = vscode.workspace.getConfiguration();
    const isDisabled = !appMod || configuration.get(`extensions.disabled.${appModExtensionId}`, false);

```

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+async function runUpgrade(promptText: string) {
```
wangmingliang-ms: why not use `Commands.JAVA_UPGRADE_WITH_COPILOT`

FluoriteCafe-work: It's used for definition of `Commands.JAVA_UPGRADE_WITH_COPILOT` - will be a loop to call that command here. Please check L37-L45.

src/upgrade/upgradeManager.ts
FluoriteCafe-work: I've checked that when the extension is disabled by user, `vscode.extensions.getExtension(appModExtensionId)` will return `undefined`. Given the test result I think the `disabled` check may be unnecessary.

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+class UpgradeManager {
+    public static initialize(context: ExtensionContext) {
+        notificationManager.initialize(context);
+
+        // Commands to be used
+        context.subscriptions.push(instrumentOperationAsVsCodeCommand(Commands.JAVA_UPGRADE_WITH_COPILOT, async (promptText?: string) => {
+            const promptToUse = promptText ?? DEFAULT_UPGRADE_PROMPT;
+            // The command should typically exist as we checked for the extension before.
+            const hasAgentModeCommand = (await commands.getCommands(true).then(cmds => cmds.includes(Commands.GOTO_AGENT_MODE)));
+            if (hasAgentModeCommand) {
```
wangmingliang-ms: no need to check if command is available here as you already checked the extension.

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+                await commands.executeCommand(Commands.GOTO_AGENT_MODE, { prompt: promptToUse });
+            } else {
+                await runUpgrade(promptToUse);
+            }
+        }));
+        commands.executeCommand('setContext', 'isModernizationExtensionInstalled',
+            !!extensions.getExtension(ExtensionName.APP_MODERNIZATION_FOR_JAVA));
```
wangmingliang-ms, line 33: also need to check if it's enabled.

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+                const projectIssues: UpgradeIssue[] = [];
+                const uri = folder.uri.toString();
+                const projects = await Jdtls.getProjects(uri);
+                await Promise.allSettled(projects.map(async (projectNode) => {
+                    const issues = await assessmentManager.getProjectIssues(projectNode);
+                    projectIssues.push(...issues);
+                    sendInfo(operationId, {
+                        issuesFoundForPackageId: projectIssues.map(x => `${x.packageId}:${x.currentVersion}`).join(","),
+                    });
+                }));
```
wangmingliang-ms: `getWorkspaceIssues`?

src/upgrade/upgradeManager.ts
FluoriteCafe-work: This is just for robustness - I feel it fine to remove the check if we can assure the command to be provided by the extension.

src/upgrade/upgradeManager.ts
FluoriteCafe-work: > I've checked that when the extension is disabled by user, `vscode.extensions.getExtension(appModExtensionId)` will return undefined. Given the test result I think the disabled check may be unnecessary. [src](https://github.com/microsoft/vscode-java-dependency/pull/901/files#r2347867584)

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                const projectIssues: UpgradeIssue[] = [];
+                const uri = folder.uri.toString();
+                const projects = await Jdtls.getProjects(uri);
+                await Promise.allSettled(projects.map(async (projectNode) => {
+                    const issues = await assessmentManager.getProjectIssues(projectNode);
+                    projectIssues.push(...issues);
+                    sendInfo(operationId, {
+                        issuesFoundForPackageId: projectIssues.map(x => `${x.packageId}:${x.currentVersion}`).join(","),
```
FluoriteCafe-work: TODO: use `JSON.stringify()`

src/upgrade/upgradeManager.ts
FluoriteCafe-work: Fixed in fa0fe9e22759f27615757d6bd6ac245c7fe6d1cd.

src/upgrade/upgradeManager.ts
FluoriteCafe-work: Removed in 1386bfb16a5269f5b7411aa1458896d217750fd0.

src/upgrade/upgradeManager.ts
FluoriteCafe-work: -> [Comment #r2348034387](https://github.com/microsoft/vscode-java-dependency/pull/901/files#r2348034387)

src/upgrade/assessmentManager.ts
FluoriteCafe-work: Refactored in 4bec02348610b94d1ff449caade464d672fb19a5.

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,73 @@
...
+        if (!shouldCheckUpgrade()) {
+            return;
+        }
+        workspace.workspaceFolders?.forEach((folder) =>
+            UpgradeManager.checkUpgradableComponents(folder)
+        );
+    }
+
+    private static async checkUpgradableComponents(folder: WorkspaceFolder) {
+        return (instrumentOperation("upgradeManager.checkUpgradableComponents",
```
wangmingliang-ms: "java.dependency.diagnoseDependencyIssues"

package.json
```diff
@@ -323,6 +333,11 @@
           "description": "%configuration.java.dependency.packagePresentation%",
           "default": "flat"
         },
+        "java.dependency.enableDependencyDiagnostics": {
```
wangmingliang-ms: `java.dependency.enableDependencyDiagnosing`?

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,73 @@
...
+    public static scan() {
+        if (!shouldCheckUpgrade()) {
+            return;
+        }
+        workspace.workspaceFolders?.forEach((folder) =>
+            UpgradeManager.checkUpgradableComponents(folder)
+        );
+    }
+
+    private static async checkUpgradableComponents(folder: WorkspaceFolder) {
```
wangmingliang-ms: `diagnoseDependencyIssues`? to be consistent with the user settings.

src/upgrade/upgradeManager.ts
FluoriteCafe-work: Fixed in 5e0f4b0d587800526f4a3f872446a65ac5fd7bc7.

package.json
FluoriteCafe-work: Renamed to `java.dependency.enableDependencyCheckup` - 9fa24f97904b4cc65d285b15cc9ff546761f42de.

src/upgrade/upgradeManager.ts
FluoriteCafe-work: -> [comment thread](https://github.com/microsoft/vscode-java-dependency/pull/901#discussion_r2348075115)

wangmingliang-ms (approved):
LGTM

wangmingliang-ms (approved):
LGTM

src/upgrade/display/notificationManager.ts
```diff
@@ -0,0 +1,90 @@
...
+                const issue = issues[0];
+
+                if (this.hasShown) {
+                    return;
+                }
+                this.hasShown = true;
+
+                if (!this.shouldShow()) {
+                    return;
+                }
```
jdneo: I think we should call `shouldShow()` check before setting the hasShown to true.

src/upgrade/upgradeManager.ts
```diff
@@ -0,0 +1,73 @@
...
+                if (!await languageServerApiManager.ready()) {
+                    return;
+                }
+                const hasJavaError: boolean = await Jdtls.checkImportStatus();
+                if (hasJavaError) {
+                    return;
+                }
+
+                const uri = folder.uri.toString();
+                const workspaceIssues = await assessmentManager.getWorkspaceIssues(uri);
```
jdneo, line 62: Assume a workspace has two workspace folders, will getWorkspaceIssues(uri) only returns the issues under that folder?

I cannot remember the details of the java side implementation so just asking for conformation

FluoriteCafe-work: This function accepts only one path. If you mean "workspace" by VSCode "workspace", it's `scan()` that get the full list of directories in the workspace via `workspace.workspaceFolders` and pass them one by one to `Jdtls`.

data in a workspace with multiple folders


<details>

<summary>note: test code used:</summary>

```js
setInterval(() => {
            console.log("Folders", workspace.workspaceFolders, workspace.workspaceFolders?.map(X => X.uri));
        }, 1000);
```

</details>

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: It's put in this order as `shouldShow()` has a bit more cost as it reads `Settings` and `globalState`. Currently the ordering seems to decide one difference - when user enable the setting entry after the initial scan and trigger a notification (e.g. by adding a directory to workspace), the ordering would affect whether a notification is shown.

I'll tweak the order to clarify the intended behavior of the variable name `hasShown`.

FluoriteCafe-work: Fixed in 478d871575deca71c404cbab22c811bb9ab76217.

src/upgrade/upgradeManager.ts
jdneo: image

The parameter name here can be changed to workspaceFolderUri.

And pls make sure Jdtls will only return projects belong to that folder

FluoriteCafe-work: > The parameter name here can be changed to workspaceFolderUri.

Fixed in 4397ccce8929635799d8b73d13f3666be931093a.

> And pls make sure Jdtls will only return projects belong to that folder

The way this file calls `Jdtls` is adopted from current `Jdtls` usage, so I think it should work as expected:

<https://github.com/microsoft/vscode-java-dependency/blob/852589ac0c3a9454829c23c780e1f843aa049498/src/views/dependencyDataProvider.ts#L191-L206>

jdneo (dismissed):

src/upgrade/utility.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    switch (reason) {
+        case UpgradeReason.CVE: {
+            return `The current project is using ${packageDisplayName} ${currentVersion}, which has CVE issues. Do you want to upgrade to ${suggestedVersionName} (${suggestedVersionDescription})?`;
+        }
+        case UpgradeReason.DEPRECATED:
+        case UpgradeReason.END_OF_LIFE:
+        default: {
+            return `The current project is using ${packageDisplayName} ${currentVersion}, which has reached end of life. Do you want to upgrade to ${suggestedVersionName} (${suggestedVersionDescription})?`;
```
akaroml: “which has reached end of life”
Can we be explicit about the time of "end of life"?

FluoriteCafe-work: Fixed by 4d271c153fe3d9d88d1376b919cb324e88ab5f1d.

The current project is using Spring Boot 3.1.3, which has reached end of life in 2025-06. [...]

jdneo (approved):

package.json
```diff
@@ -323,6 +333,11 @@
           "description": "%configuration.java.dependency.packagePresentation%",
           "default": "flat"
         },
+        "java.dependency.enableDependencyCheckup": {
+          "type": "boolean",
+          "description": "%configuration.java.dependency.enableDependencyCheckup%",
+          "default": true
+        },
```
linghengqian, line 340: I would say I'm confused as to why the documentation for `java.dependency.enableDependencyCheckup` isn't updated in https://github.com/microsoft/vscode-java-dependency/blob/main/README.md . This at least affects https://github.com/microsoft/vscode-java-dependency/issues/917 .

FluoriteCafe-work: Thanks for the reminder! We've added it in #958.

PR #907, opened 2025-09-18
feat: add getImportClassContent method to collect import class info

wenytang-ms:
Testing against some popular java repo 
[jbang](https://github.com/jbangdev/jbang)
[Mixin](https://github.com/SpongePowered/Mixin)
[Velocity](https://github.com/PaperMC/Velocity)
java deps cost - execute java command time cost
pack cost - execute context provider to collect and return context whole time cost

|file url|java-deps cost|pack cost|
|--|--|--|
|[Configuration.java](https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/Configuration.java)|9.82 ms|13 ms 7 items|
|[PathTypeAdapter.java](https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/util/PathTypeAdapter.java)|4.36ms |15ms 0 items|
|[VersionChecker.java](https://github.com/jbangdev/jbang/blob/main/src/main/java/dev/jbang/util/VersionChecker.java)|4.6ms|13ms 3 items|
|[RuntimeDecompiler.java](https://github.com/SpongePowered/Mixin/blob/master/src/fernflower/java/org/spongepowered/asm/mixin/transformer/debug/RuntimeDecompiler.java)|20ms |25ms 4 items|
|[ModLauncherClassTracker.java](https://github.com/SpongePowered/Mixin/blob/master/src/modlauncher/java/org/spongepowered/asm/service/modlauncher/ModLauncherClassTracker.java)|7.74ms|19ms 3items|
|[CookieReceiveEvent.java](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/api/src/main/java/com/velocitypowered/api/event/player/CookieReceiveEvent.java)|4.5ms|8ms 6 items|
|[ResultEvent.java](https://github.com/PaperMC/Velocity/blob/dev/3.0.0/api/src/main/java/com/velocitypowered/api/event/ResultedEvent.java)|5.2 ms| 12 ms 0 items|

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
@@ -42,7 +42,6 @@ import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.MultiStatus;
-import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.jdt.core.IJavaElement;
@@ -66,7 +65,6 @@ import org.eclipse.jdt.ls.core.internal.ResourceUtils;
 import org.eclipse.jdt.ls.core.internal.managers.ProjectsManager;
 import org.eclipse.jdt.ls.core.internal.managers.UpdateClasspathJob;
 import org.eclipse.jdt.ls.core.internal.preferences.Preferences.ReferencedLibraries;
-import org.eclipse.jdt.ls.core.internal.preferences.Preferences.SearchScope;
 import org.eclipse.lsp4j.jsonrpc.json.adapters.CollectionTypeAdapter;
 import org.eclipse.lsp4j.jsonrpc.json.adapters.EnumTypeAdapter;
 
@@ -85,6 +83,16 @@ public final class ProjectCommand {
         public MainClassInfo(String name, String path) {
             this.name = name;
             this.path = path;
+        }
+    }
+
+    private static class ImportClassInfo {
+        public String uri;
+        public String className;
+
+        public ImportClassInfo(String uri, String className) {
+            this.uri = uri;
+            this.className = className;
         }
     }
 
@@ -110,7 +118,7 @@ public final class ProjectCommand {
             projects = ProjectUtils.getAllProjects();
         } else {
             projects = Arrays.stream(ProjectUtils.getJavaProjects())
-                .map(IJavaProject::getProject).toArray(IProject[]::new);
+                    .map(IJavaProject::getProject).toArray(IProject[]::new);
         }
 
         ArrayList<PackageNode> children = new ArrayList<>();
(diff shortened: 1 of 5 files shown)


jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
```diff
@@ -332,11 +348,282 @@ public static boolean checkImportStatus() {
...
+    private static void resolveSingleType(IJavaProject javaProject, String typeName, List<ImportClassInfo> result,
+            Set<String> processedTypes) {
+        try {
+            if (processedTypes.contains(typeName)) {
+                return;
+            }
+            processedTypes.add(typeName);
+
+            // Find the type in the project
+            org.eclipse.jdt.core.IType type = javaProject.findType(typeName);
```
testforstephen: findType is an expensive operation, could you resolve the type node from IImportDeclaration directly?

wenytang-ms: updated!

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
```diff
@@ -332,11 +348,331 @@ public static boolean checkImportStatus() {
...
+                }
+
+                String importName = importDecl.getElementName();
+                if (importName.endsWith(".*")) {
+                    // Handle package imports
+                    String packageName = importName.substring(0, importName.length() - 2);
+                    resolvePackageTypes(javaProject, packageName, result, processedTypes);
+                } else {
+                    // Handle single type imports
+                    resolveSingleType(javaProject, importName, result, processedTypes);
```
testforstephen, line 413: What I mean is that you can infer the type object from `org.eclipse.jdt.core.IImportDeclaration importDecl` model directly.

Also you need to consider static import case as well.

wenytang-ms: the `IImportDeclaration` only provide the name of the import package name, no other type info. 
Both check with copilot and src code https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IImportDeclaration.java

wenytang-ms: also add two method to support static import case

testforstephen: > the `IImportDeclaration` only provide the name of the import package name, no other type info. Both check with copilot and src code https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IImportDeclaration.java

I don't think so. Do you debug the memory variable? I believe it has more info on the impl object.

wenytang-ms: IImportDeclaration only return current editing file instead of the import class. so can't get the import class info

testforstephen (approved):
This is a good start to enable the exp first.

There are more cases to cover for the import, we can optimize them later. for example, In Java, all classes in the same package can access each other by default without explicit imports. Also can support the method with package visible or protected modifier.

wenytang-ms:
Thanks @testforstephen for your suggestion, will update the logic for better user experiment.

This PR will be merged after the stable release of the Java mod change.

PR #910, opened 2025-09-25
Calculate package version range for upgrade

FluoriteCafe-work:


src/upgrade/display/notificationManager.ts
```diff
@@ -53,6 +53,10 @@ class NotificationManager implements IUpgradeIssuesRenderer {
                 const prompt = buildFixPrompt(issue);
                 const notificationMessage = buildNotificationMessage(issue, hasExtension);
                 const upgradeButtonText = hasExtension ? BUTTON_TEXT_UPGRADE : BUTTON_TEXT_INSTALL_AND_UPGRADE;
+
+                sendInfo(operationId, {
```
Flanker32, line 57: may move this line later to track users selection

also we may track info like whether users has extension installed, the total num of dependencies issues....

FluoriteCafe-work: > may move this line later to track users selection

Line 65 `java.dependency.upgradeNotification.runUpgrade` handles that. This event is for the case when user does not act on the notification (i.e. neither clicking any button nor dismissing the notification).

> also we may track info like whether users has extension installed, the total num of dependencies issues....

This may make sense now that we don't have AppMod in the pack.

src/upgrade/assessmentManager.ts
```diff
@@ -84,7 +97,20 @@ async function getDependencyIssues(projectNode: INodeData): Promise<UpgradeIssue
...
 
-                return packages.map(getDependencyIssue).filter((x): x is UpgradeIssue => Boolean(x));
+                const versionsByGroupId: VersionSet = {};
+                const issues = packages.map(pkg => getDependencyIssue(pkg, versionsByGroupId)).filter((x): x is UpgradeIssue => Boolean(x));
+                if (Object.entries(versionsByGroupId).length > 0) {
+                    sendInfo("", {
+                        operationName: "java.dependency.assessmentManager.getDependencyIssues",
+                        versionRangeByGroupId: JSON.stringify(
+                            Object.fromEntries(
+                                Object.entries(versionsByGroupId).map(([groupId, versions]) => [groupId, getVersionRange(versions)]),
```
wangmingliang-ms: this will send the versions of all groupIds to Microsoft, not allowed.

src/upgrade/display/notificationManager.ts
```diff
@@ -53,6 +53,10 @@ class NotificationManager implements IUpgradeIssuesRenderer {
...
+                    operationName: "java.dependency.upgradeNotification.runUpgradeNotification",
```
wangmingliang-ms:
```suggestion
                    operationName: "java.dependency.upgradeNotification.show",
```

src/upgrade/assessmentManager.ts
```diff
@@ -84,7 +97,20 @@ async function getDependencyIssues(projectNode: INodeData): Promise<UpgradeIssue
                     path: packageContainer.path,
                 });
 
-                return packages.map(getDependencyIssue).filter((x): x is UpgradeIssue => Boolean(x));
+                const versionsByGroupId: VersionSet = {};
+                const issues = packages.map(pkg => getDependencyIssue(pkg, versionsByGroupId)).filter((x): x is UpgradeIssue => Boolean(x));
```
wangmingliang-ms: bad design to pass "versionsByGroupId" to getDependencyIssue.
can you make "versions" a property of UpgradeIssue?

src/upgrade/assessmentManager.ts
FluoriteCafe-work: Absence of [`supportedVersionDefinition`](https://github.com/FluoriteCafe-work/vscode-java-dependency/blob/a81ad083b79605a8738d9495899b4611ee6cdf8e/src/upgrade/assessmentManager.ts#L75-L76) will prevent packages with a `groupId` outside the metadata to be added to telemetry. Therefore, it's still the case that packages on which we check the version are the only ones that get included in telemetry.

This may not seem clear enough - do you think I need to do some refactoring or comments to stress that?

src/upgrade/assessmentManager.ts
FluoriteCafe-work: I also agree it's not looking right indeed. Made some changes to split the two package check flows and cleaned it up: 44a0b06.

src/upgrade/display/notificationManager.ts
FluoriteCafe-work: > also we may track info like whether users has extension installed, the total num of dependencies issues....

@Flanker32 do you think we should put such data in some startup/activation telemetry?

PR #919, opened 2025-10-23
feat: add project deps info to context

wenytang-ms:
policy updated
1. get class info aim to 70ms return. internal class first, then external
2. add dependency info interface, to list project deps, get current project java version, maven version and so on.

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
@@ -350,9 +350,6 @@ public final class ProjectCommand {
         return hasError;
     }
 
-    // Maximum methods to display for binary (external) classes
-    private static final int MAX_METHODS_FOR_BINARY = 50;
-
     /**
      * Get import class content for Copilot integration.
      * This method extracts information about imported classes from a Java file.
@@ -464,7 +461,7 @@ public final class ProjectCommand {
                         // Resolve external (binary) types with simplified content
                         if (!isStatic) {
                             ContextResolver.resolveBinaryType(javaProject, importName, externalClasses, 
-                                    processedTypes, MAX_METHODS_FOR_BINARY, monitor);
+                                    processedTypes, Integer.MAX_VALUE, monitor);
                         }
                     }
                     


jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
```diff
@@ -341,9 +350,20 @@ public static boolean checkImportStatus() {
...
 
+    // Threshold for triggering external dependency resolution
+    // If project source classes < this value, supplement with external dependencies
+    private static final int MIN_SOURCE_CLASSES_THRESHOLD = 5;
+    
+    // Maximum number of external dependency classes to include
+    private static final int MAX_DEPENDENCY_CLASSES = 10;
+    
+    // Maximum methods to display for binary (external) classes
+    private static final int MAX_METHODS_FOR_BINARY = 5;
```
testforstephen: who do we need define these magic numbers to limit the context number?

wenytang-ms: For the class info of external dependency resolution, we take into account cases where there are too many methods, so we limit the extraction to 5.

wenytang-ms: let me cancel such limitation

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
```diff
@@ -416,6 +437,36 @@ public static List<ImportClassInfo> getImportClassContent(List<Object> arguments
                 }
             }
 
+            // Phase 2: Adaptive external dependency resolution
+            // Only trigger if project source classes are sparse (< threshold)
+            if (classInfoList.size() < MIN_SOURCE_CLASSES_THRESHOLD && !monitor.isCanceled()) {
```
chagong: now the strategy is, when the num of internal dependencies is less than 5 we add external dependencies info? And I can see now you only take totally 10 dependencies? I agree that we should take internal dependencies higher priority than external dependencies, why don't we just limit the total number and handle internal dependencies first?

wenytang-ms: yes, we handle internal deps first, without any num limit. we only limit to external deps.
currently we don't limit to internal deps, directly limiting the total number would restrict internal classes, which is not the intended behavior.

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
testforstephen: I don't like setting the threshold at the beginning without any data support. You can add BI track to collect the total class, method number and time cost for each context request. If limiting the number can improve performance, then we can do it.

wenytang-ms: this solution is work for prerelease, because this PR is going to merge into stable, so I think we can loose the limitation on stable, and cancel such limitation on prerelease.

testforstephen: One idea to truncate the import context list dynamically based on the time budget. For example, grant 75 ms to resolve import source classes and 50 ms to resolve external classes. Since you use a for loop to resolve the import list, you can determine whether to break the for-loop earlier based on the remaining time budget.

testforstephen (approved):

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
wenytang-ms: update.

jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
wenytang-ms: policy updated

PR #924, opened 2025-10-30
feat: support context provider api

wenytang-ms:
feat: add context provider to vscode java dependency because exp support filter java manager

### DIFF
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/ProjectCommand.java
@@ -506,7 +506,7 @@ public final class ProjectCommand {
     /**
      * Get project dependencies information including JDK version.
      * 
-     * @param arguments List containing the project URI as the first element
+     * @param arguments List containing the file URI as the first element
      * @param monitor Progress monitor for cancellation support
      * @return List of DependencyInfo containing key-value pairs of project information
      */
@@ -515,8 +515,8 @@ public final class ProjectCommand {
             return new ArrayList<>();
         }
 
-        String projectUri = (String) arguments.get(0);
-        List<ProjectResolver.DependencyInfo> resolverResult = ProjectResolver.resolveProjectDependencies(projectUri, monitor);
+        String fileUri = (String) arguments.get(0);
+        List<ProjectResolver.DependencyInfo> resolverResult = ProjectResolver.resolveProjectDependencies(fileUri, monitor);
         
         // Convert ProjectResolver.DependencyInfo to ProjectCommand.DependencyInfo
         List<DependencyInfo> result = new ArrayList<>();
--- a/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/parser/ProjectResolver.java
+++ b/jdtls.ext/com.microsoft.jdtls.ext.core/src/com/microsoft/jdtls/ext/core/parser/ProjectResolver.java
@@ -2,13 +2,25 @@ package com.microsoft.jdtls.ext.core.parser;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IResourceChangeEvent;
+import org.eclipse.core.resources.IResourceChangeListener;
+import org.eclipse.core.resources.IResourceDelta;
+import org.eclipse.core.resources.IResourceDeltaVisitor;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.ElementChangedEvent;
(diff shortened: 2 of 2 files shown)


src/copilot/contextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+            selector: [{ language: "java" }],
+            resolver: { resolve: createJavaContextResolver() }
+        };
+        const installCount = await JavaContextProviderUtils.installContextProviderOnApis(apis, provider, context, installContextProvider);
+        if (installCount === 0) {
+            return;
+        }
+        sendInfo("", {
+            "action": "registerCopilotContextProvider",
+            "extension": 'vscjava.vscode-java-dependency',
```
testforstephen: this is not needed. The telemetry library already includes the extension id info in all telemetry event.

src/copilot/contextProvider.ts
```diff
@@ -0,0 +1,154 @@
...
+        if (!apis.clientApi || !apis.chatApi) {
+            return;
+        }
+        // Register the Java completion context provider
+        const provider: ContextProvider<SupportedContextItem> = {
+            id: 'vscjava.vscode-java-dependency', // use extension id as provider id for now
+            selector: [{ language: "java" }],
+            resolver: { resolve: createJavaContextResolver() }
+        };
+        const installCount = await JavaContextProviderUtils.installContextProviderOnApis(apis, provider, context, installContextProvider);
```
testforstephen, line 38: what will you do to the context provider on vscode-java-pack extension?

src/copilot/copilotHelper.ts
```diff
@@ -0,0 +1,358 @@
...
+
+    /**
+     * Resolves project dependencies and converts them to context items with cancellation support
+     * @param workspaceFolders The workspace folders, or undefined if none
+     * @param copilotCancel Cancellation token from Copilot
+     * @param checkCancellation Function to check for cancellation
+     * @returns Array of context items for project dependencies, or empty array if no workspace folders
+     */
+    export async function resolveAndConvertProjectDependencies(
+        workspaceFolders: readonly { uri: Uri }[] | undefined,
```
testforstephen: A workspace folder is not necessarily mapped to a Java project. A better approach is to find the related Java project based on the active editor file, not based on the workspace folder info.

src/copilot/contextProvider.ts
wenytang-ms: I will remove context provider in the next version of the Java pack. 
As long as the feature flag is not enabled, it will not affect the Java pack.

src/copilot/utils.ts
```diff
@@ -0,0 +1,233 @@
...
+
+        return installCount;
+    }
+
+    /**
+     * Calculate approximate token count for context items
+     * Using a simple heuristic: ~4 characters per token
+     * Optimized for performance by using reduce and direct property access
+     */
+    static calculateTokenCount(items: SupportedContextItem[]): number {
```
chagong, line 118: why we need this? does copilot side have this kind of token count which could be more accurate?

src/copilot/copilotHelper.ts
```diff
@@ -0,0 +1,358 @@
...
+        copilotCancel: CancellationToken,
+        checkCancellation: (token: CancellationToken) => void
+    ): Promise<{ name: string; value: string; importance: number }[]> {
+        const items: any[] = [];
+        // Check if workspace folders exist
+        if (!workspaceFolders || workspaceFolders.length === 0) {
+            sendContextOperationTelemetry("resolveProjectDependencies", "ContextEmpty", sendInfo, EmptyReason.NoWorkspace);
+            return items;
+        }
+        const projectUri = workspaceFolders[0];
```
chagong: not sure in what cases that workspaceFolders.length is bigger than 1. But if it's more than 1 items in workspaceFolders, is it OK that you choose the [0] item as project URI?

src/copilot/copilotHelper.ts
wenytang-ms: yes, this is a good point! I also find that java project have aggregate project, so this PR is going to support such scenario
https://github.com/microsoft/vscode-java-dependency/pull/926
My consideration is that if, in the end, we can provide the entire workspace's project info, then we can consider updating them simultaneously.

src/copilot/utils.ts
wenytang-ms: This is different because Copilot does not maintain separate token tracking for the context we provide; it treats it as part of the overall input.
We do really need such for a better understanding to our context

src/copilot/copilotHelper.ts
wenytang-ms: Your point makes sense, and it’s possible that some users work with multiple workspaces in VS Code. However, I don’t think we need to rush a solution now. Let’s first collect some data to understand how common this scenario is. My guess is that the proportion won’t be high.

src/copilot/copilotHelper.ts
testforstephen: My question is that the completion provider is targeted for a opened Java file, why do you pass the workspace folders here instead of the file uri? In jdt.ls side, you can find the associated Java project from the Java compilation unit directly.

src/copilot/contextProvider.ts
wenytang-ms: removed

src/copilot/copilotHelper.ts
wenytang-ms: updated, will pass the file uri into java delegate, and handle the project info in the java part. #926

src/copilot/copilotHelper.ts
wenytang-ms: update this PR and now pass the file uri instead of project uri
Java part will handle such uri to the right java project

mahdidevil8:
Great

mahdidevil8:
Great

mahdidevil8 (approved):

PR #947, opened 2025-11-18
Always show Java projects even when errors occur during import

### LINKED ISSUE #946: "Oops..." Message instead of Java Project even though there are detected Projects
I have a project with many maven project that imports with project errors. The problem I have is that the "Java Projects" view remains empty and shows a message starting with "Oops, something went wrong when opening Java projects...." even though projects some Java projects are being detected. This makes it hard to determine which projects import just fine and which do not. We should always show projects that have been detected.

tsmaeder:
Fixes #946.


tsmaeder:
The CI failures don't look like something this PR could cause 🤷

PR #948, opened 2025-12-03
feat: add CVE checking in java upgrade promotion

yezhu6:
Add critical and high severity CVE issues checking.
image

### DIFF
--- a/src/upgrade/assessmentManager.ts
+++ b/src/upgrade/assessmentManager.ts
@@ -183,17 +183,15 @@ async function getAllDependencies(projectNode: INodeData): Promise<PackageDescri
         })
     );
 
-    return allPackages
-        .map(x => {
-            if (x.status === "fulfilled") {
-                return x.value;
-            }
-            sendInfo("", {
-                operationName: "java.dependency.assessmentManager.getAllDependencies.packageDataFailure",
-            });
-            return [];
-        })
-        .flat();
+    const fulfilled = allPackages.filter((x): x is PromiseFulfilledResult<PackageDescription[]> => x.status === "fulfilled");
+    const failedPackageCount = allPackages.length - fulfilled.length;
+    if (failedPackageCount > 0) {
+        sendInfo("", {
+            operationName: "java.dependency.assessmentManager.getAllDependencies.rejected",
+            failedPackageCount: String(failedPackageCount),
+        });
+    }
+    return fulfilled.map(x => x.value).flat();
 }
 
 async function getCVEIssues(dependencies: PackageDescription[]): Promise<UpgradeIssue[]> {


src/upgrade/utility.ts
```diff
@@ -50,7 +51,36 @@ export function buildNotificationMessage(issue: UpgradeIssue, hasExtension: bool
...
+    if (highCount > 0) {
+        parts.push(`${highCount} high-severity`);
+    }
+
+    const severityText = parts.join(' and ');
+    if (issues.length === 1) {
+        return `${severityText} CVE vulnerability is detected in this project. Would you like to fix it now?`;
+    }
+
+    return `${severityText} CVE vulnerabilities are detected in this project. Would you like to fix them now?`;
```
FluoriteCafe-work: nit: the command `JAVA_UPGRADE_WITH_COPILOT` will install AppMod extension (if not found). You may want to tell user about the behavior in the message.

src/upgrade/assessmentManager.ts
```diff
@@ -122,50 +123,32 @@ function getDependencyIssue(pkg: PackageDescription): UpgradeIssue | null {
...
-        .reduce((a, b) => [...a, ...b]);
+    return issues;
 }
 
 async function getProjectIssues(projectNode: INodeData): Promise<UpgradeIssue[]> {
-    const issues: UpgradeIssue[] = [];
-    issues.push(...getJavaIssues(projectNode));
-    issues.push(...(await getDependencyIssues(projectNode)));
-    return issues;
+    const cveIssues = await getCVEIssues(projectNode);
```
FluoriteCafe-work: Please reuse the dependency list from `getAllDependencies()` in `getDependencyIssues()` and `getCVEIssues()`. Calling `Jdtls` can be a bit expensive.

src/upgrade/assessmentManager.ts
```diff
@@ -122,50 +123,32 @@ function getDependencyIssue(pkg: PackageDescription): UpgradeIssue | null {
...
+    if (cveIssues.length > 0) {
+        return cveIssues;
```
FluoriteCafe-work: Please don't early-return here as it may cause issues in later data processing steps.

src/upgrade/display/notificationManager.ts
```diff
@@ -51,22 +54,45 @@ class NotificationManager implements IUpgradeIssuesRenderer {
...
+                    // Filter to only CVE issues and cast to CveUpgradeIssue[]
+                    cveIssues = issues.filter(
+                        (i): i is CveUpgradeIssue => i.reason === UpgradeReason.CVE
+                    );
+                    // Sort by severity (critical first, then high, etc.)
+                    cveIssues.sort((a, b) => {
+                        const severityA = (a.severity?.toLowerCase().trim() || 'unknown') as Severity;
+                        const severityB = (b.severity?.toLowerCase().trim() || 'unknown') as Severity;
+                        return (severityOrder[severityB] ?? 0) - (severityOrder[severityA] ?? 0);
+                    });
```
wangmingliang-ms: can be simplified if use numeric enum.

src/upgrade/cve.ts
```diff
@@ -0,0 +1,154 @@
...
+ */
+export const severityOrder = {
+    critical: 4,
+    high: 3,
+    medium: 2,
+    low: 1,
+    unknown: 0,
+} as const;
+
+export type Severity = keyof typeof severityOrder;
```
wangmingliang-ms: may be simplified using numeric enum
https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums

src/upgrade/cve.ts
```diff
@@ -0,0 +1,154 @@
...
+                link: mostCriticalCve.html_url,
+            };
+        });
+        return upgradeIssues;
+    } catch (error) {
+        throw error;
+    }
+}
+
+function findMostCriticalCve(depCves: CVE[]) {
```
wangmingliang-ms: sort and get the first one, may make code clearer.

src/upgrade/cve.ts
```diff
@@ -0,0 +1,154 @@
...
+                reason: UpgradeReason.CVE as const,
+                suggestedVersion: {
+                    name: depCve.minVersion || 'unknown',
+                    description: mostCriticalCve.description || mostCriticalCve.summary || 'Security vulnerability detected'
+                },
+                severity: mostCriticalCve.severity,
+                description: mostCriticalCve.description || mostCriticalCve.summary || 'Security vulnerability detected',
+                link: mostCriticalCve.html_url,
+            };
+        });
```
wangmingliang-ms: please split the logic of getting cves and converting them into UpgradeIssue-s into 2 methods.

src/upgrade/utility.ts
yezhu6: image

src/upgrade/display/notificationManager.ts
```diff
@@ -2,18 +2,21 @@
...
 import { ExtensionName } from "../../constants";
+import { CveUpgradeIssue } from "../cve";
 
 const KEY_PREFIX = 'javaupgrade.notificationManager';
 const NEXT_SHOW_TS_KEY = `${KEY_PREFIX}.nextShowTs`;
 
 const BUTTON_TEXT_UPGRADE = "Upgrade Now";
+const BUTTON_TEXT_FIX_CVE = "Fix CVE Issues";
 const BUTTON_TEXT_INSTALL_AND_UPGRADE = "Install Extension and Upgrade";
+const BUTTON_TEXT_INSTALL_AND_FIX_CVE = "Install Extension and Fix CVE Issues";
```
wangmingliang-ms:
```suggestion
const BUTTON_TEXT_INSTALL_AND_FIX_CVE = "Install Extension and Fix";
```

src/upgrade/display/notificationManager.ts
```diff
@@ -2,18 +2,21 @@
...
 import { Settings } from "../../settings";
 import { instrumentOperation, sendInfo } from "vscode-extension-telemetry-wrapper";
 import { ExtensionName } from "../../constants";
+import { CveUpgradeIssue } from "../cve";
 
 const KEY_PREFIX = 'javaupgrade.notificationManager';
 const NEXT_SHOW_TS_KEY = `${KEY_PREFIX}.nextShowTs`;
 
 const BUTTON_TEXT_UPGRADE = "Upgrade Now";
+const BUTTON_TEXT_FIX_CVE = "Fix CVE Issues";
```
wangmingliang-ms:
```suggestion
const BUTTON_TEXT_FIX_CVE = "Fix Now";
```

src/upgrade/cve.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+function calculateMaxPatchedVersion(
+  depCves: CVE[],
+  dep: { name: string; version: string }
+) {
+  let maxPatchedVersion: string | undefined | null;
+
+  for (const cve of depCves) {
+    const patchedVersion = cve.affectedDeps.find(
```
wangmingliang-ms: we don't need patch version here, right?

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to the "vscode-java-dependency" extension will be documented
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 0.27.0
+
+- feat  - Add CVE checking to nodify users to fix the critical/high-severity CVE issues by yezhu6 in https://github.com/microsoft/vscode-java-dependency/pull/948
```
wangmingliang-ms:
```suggestion
- feat  - Add CVE checking to nodify users to fix the critical/high-severity CVE issues in https://github.com/microsoft/vscode-java-dependency/pull/948
```

wangmingliang-ms (approved):
LGTM

src/upgrade/assessmentManager.ts
```diff
@@ -184,11 +163,44 @@ async function getWorkspaceIssues(workspaceFolderUri: string): Promise<UpgradeIs
...
+            return packageNodes.map(packageNodeToDescription).filter((x): x is PackageDescription => Boolean(x));
+        })
+    );
+
+    return allPackages
+        .map(x => {
+            if (x.status === "fulfilled") {
+                return x.value;
+            }
+            sendInfo("", {
```
chagong: I won't suggest send telemetry here in map. In case of big project that contains hundreds/thousands of packages, you may send big amount of telemetry event in a very short time, and causes telemetry data loss. 

I would suggest set telemetry properties here and send only one event in the end.

src/upgrade/cve.ts
```diff
@@ -0,0 +1,188 @@
...
+    return [];
+  }
+  const octokit = new Octokit();
+
+  const response = await octokit.securityAdvisories.listGlobalAdvisories({
+    ecosystem: "maven",
+    affects: deps.map((p) => `${p.name}@${p.version}`),
+    direction: "asc",
+    sort: "published",
+    per_page: 100,
```
chagong: what about the listed advisories is larger than 100? You don't handle pagination?

src/upgrade/assessmentManager.ts
yezhu6: In practice, allPackages is unlikely to have thousands of items because:

1. It's not counting individual dependencies - allPackages contains one entry per package container (like "Maven Dependencies", "Referenced Libraries", etc.), not per individual dependency.

2. Typical Java projects have very few containers - Usually just 2-5 containers per project (e.g., Maven Dependencies, JRE System Library, Project and External Dependencies).

3. The thousands of dependencies are nested inside - Each container result (x.value) contains an array of PackageDescription objects. Those inner arrays can potentially have hundreds of dependencies, but the outer allPackages array itself will be small.

PR #952, opened 2025-12-16
Get the direct dependencies while checking for promotion

yezhu6:
1.  add the tool id into the java upgrade prompts to specify the calling tool
2. Change the getDependencies to getDirectDependencies to exclude the transitive dependencies

### DIFF
--- a/src/upgrade/assessmentManager.ts
+++ b/src/upgrade/assessmentManager.ts
@@ -3,7 +3,7 @@
 
 import * as fs from 'fs';
 import * as semver from 'semver';
-import {globby} from 'globby';
+import { globby } from 'globby';
 
 import { Uri } from 'vscode';
 import { Jdtls } from "../java/jdtls";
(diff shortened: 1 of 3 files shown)


src/upgrade/assessmentManager.ts
```diff
@@ -175,30 +178,181 @@ async function getWorkspaceIssues(workspaceFolderUri: string): Promise<UpgradeIs
...
+            const groupId = match[1].trim();
+            const artifactId = match[2].trim();
+            // Skip property references like ${project.groupId}
+            if (!groupId.includes('${') && !artifactId.includes('${')) {
+                directDeps.add(`${groupId}:${artifactId}`);
+            }
+        }
+
+        // Check for parent pom in multi-module projects
+        const parentPomPath = path.join(path.dirname(pomPath), '..', 'pom.xml');
```
wangmingliang-ms: this may not be the fact. I think you can find all the pom.xml in this project first and then parse them one by one to find all the dependencies.

src/upgrade/assessmentManager.ts
```diff
@@ -175,30 +178,181 @@ async function getWorkspaceIssues(workspaceFolderUri: string): Promise<UpgradeIs
...
+    } catch {
+        // If we can't read the gradle file, return empty set
+    }
+    return directDeps;
+}
+
+/**
+ * Find the build file (pom.xml or build.gradle) for a project
+ */
+function findBuildFile(projectUri: string | undefined): { path: string; type: 'maven' | 'gradle' } | null {
```
wangmingliang-ms: you can use "glob" to find the pom.xml files, remember to exclude certain folders to improve performance. you can find the usage of glob in vscode-java-upgrade.

chagong (dismissed):

wangmingliang-ms (dismissed):

chagong (dismissed):

chagong (approved):

PR #982, opened 2026-03-27
perf: progressive project tree view during import

chagong:
## Summary

Show Java project names in the JAVA PROJECTS tree view progressively as they are imported, instead of waiting for the entire import to complete. This reduces perceived loading time for large projects (e.g., 436 Gradle subprojects) from ~7 minutes to ~1 minute.

## Problem

For large Gradle projects, the JDTLS import takes 6-7 minutes. During this time, the JAVA PROJECTS tree view is completely empty because:
1. `serverReady()` doesn't resolve until import completes
2. Even if we wait for `serverRunning()`, the server is blocked by Eclipse workspace locks and can't respond to queries like `java.project.list`

## Solution

Instead of querying the server for project data, create tree nodes directly from `ProjectsImported` notification URIs sent progressively by JDTLS during import.

#### `languageServerApiManager.ts`
- Use `serverRunning()` (API >= 0.14) instead of `serverReady()` so the tree view can start rendering before import finishes
- Route `onDidProjectsImport` events to `addProgressiveProjects()` instead of refresh (which would query the blocked server)
- Route `onDidClasspathUpdate` during import to `addProgressiveProjects()` (non-destructive)
- After import completes (`serverReady()`), trigger full refresh to replace placeholders with real data
- Added `isFullyReady()` method

#### `dependencyDataProvider.ts`
- Added `addProgressiveProjects()` — creates `ProjectNode` items from URIs without server queries
- Guard `getChildren()` from entering `getRootNodes()` during progressive loading (prevents blocking on server queries that hang for the entire import)
- Keep TreeView progress spinner visible until first progressive items arrive via a deferred promise

#### `commands.ts`
- Added `VIEW_PACKAGE_INTERNAL_ADD_PROJECTS` command constant

## Context

This is part of a 3-repo change for progressive project loading:
1. **eclipse-jdtls/eclipse.jdt.ls#3744** — Server-side progressive notifications
2. **redhat-developer/vscode-java#4372** — `serverRunning()` API (v0.14)
3. **This PR** — Client-side progressive tree view rendering

## Testing

Tested with spring-boot project (436 Gradle subprojects):
- **Before**: Tree view empty for ~7 minutes until import completes
- **After**: 436 project names appear within ~1 minute of opening, with progress spinner showing until items arrive
- Works both with cached workspace (instant) and after "Clean Java Language Server Workspace" (progressive)
- After import completes, full refresh replaces placeholder items with proper expandable project nodes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5759,9 +5759,9 @@
       "dev": true
     },
     "node_modules/picomatch": {
-      "version": "2.3.1",
-      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.3.1.tgz",
-      "integrity": "sha512-JU3teHTNjmE2VCGFzuY8EXzCDVwEqB2a8fsIvwaStHhAWJEeVd1o1QD80CU6+ZdEXXSLbSsuLwJjkCBWqRQUVA==",
+      "version": "2.3.2",
+      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.3.2.tgz",
+      "integrity": "sha512-V7+vQEJ06Z+c5tSye8S+nHUfI51xoXIXjHQ99cQtKUkQqqO1kO/KCJUfZXuB47h/YBlDhah2H3hdUGXn8ie0oA==",
       "engines": {
         "node": ">=8.6"
       },
@@ -12667,9 +12667,9 @@
       "dev": true
     },
     "picomatch": {
-      "version": "2.3.1",
-      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.3.1.tgz",
-      "integrity": "sha512-JU3teHTNjmE2VCGFzuY8EXzCDVwEqB2a8fsIvwaStHhAWJEeVd1o1QD80CU6+ZdEXXSLbSsuLwJjkCBWqRQUVA=="
+      "version": "2.3.2",
+      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.3.2.tgz",
+      "integrity": "sha512-V7+vQEJ06Z+c5tSye8S+nHUfI51xoXIXjHQ99cQtKUkQqqO1kO/KCJUfZXuB47h/YBlDhah2H3hdUGXn8ie0oA=="
     },
     "pkg-dir": {
       "version": "4.2.0",


src/views/dependencyDataProvider.ts
```diff
@@ -117,10 +122,39 @@ export class DependencyDataProvider implements TreeDataProvider<ExplorerNode> {
...
+                    this._progressiveItemsReady = new Promise<void>((resolve) => {
+                        this._resolveProgressiveItems = resolve;
+                    });
+                }
+                // Race with a timeout to prevent hanging indefinitely if no
+                // progressive notifications arrive (e.g., small project or
+                // server failure). After timeout, fall through to normal path.
+                const timeout = new Promise<void>((resolve) => setTimeout(resolve, 30_000));
+                await Promise.race([this._progressiveItemsReady, timeout]);
+            }
```
wenytang-ms: When only serverRunning() is available (vscode-java #4372 merged) but progressive ProjectsImported notifications are not (eclipse.jdt.ls #3744 not yet merged), [getChildren()](vscode-file://vscode-app/c:/Users/wenyutang/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) enters the progressive loading branch, waits 30 seconds for notifications that never arrive, then returns [this._rootItems || []](vscode-file://vscode-app/c:/Users/wenyutang/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — an empty array. This leaves the tree view blank for 30 seconds, which is worse than the current behavior (showing a loading spinner until the server is ready).

Suggested fix: After the 30-second timeout, instead of returning the empty array, fall through to the normal [getRootNodes()](vscode-file://vscode-app/c:/Users/wenyutang/AppData/Local/Programs/Microsoft%20VS%20Code/cfbea10c5f/resources/app/out/vs/code/electron-browser/workbench/workbench.html) path. This way, if no progressive items arrived, the code behaves exactly like it does today — it blocks on the server query until the server is ready, keeping the spinner visible the whole time.

```ts
if (this._rootItems && this._rootItems.length > 0) {
    return this._rootItems;
}
```
This ensures the worst case (missing external dependencies) is equivalent to today's behavior, never worse.

wenytang-ms (approved):

PR #993, opened 2026-04-22
fix: support Unicode identifiers in Java class name validation (#789)

### LINKED ISSUE #789: "New > Java Class" wizard cannot create a class for Japanese identifiers.
Originally reported from https://github.com/redhat-developer/vscode-java/issues/3289

When I open the "New > Java Class" wizard and enter a Japanese class name, I receive the message "Invalid Java qualified name." and I am unable to create a Java class.

##### Environment
- Operating System: Windows 11
- JDK version: 17
- Visual Studio Code version: 1.79.2
- Java extension version: v0.25.14

##### Steps To Reproduce
1. Select any package, right-click, and choose "New > Java Class."
2. Enter Japanese characters as the class name. For example, "ほげ" 

##### Current Result

It displays "Invalid Java qualified name."

##### Expected Result

The class is generated.

##### Additional Informations

In Java, we can use characters from languages such as Japanese and others in identifiers.

https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.8

wenytang-ms:
Change isJavaIdentifier() regex from ASCII-only /^([a-zA-Z_$][a-zA-Z\d_$]*)$/ to Unicode-aware regex using ES2018 property escapes per JLS §3.8.

This allows non-ASCII identifiers (e.g., Japanese ほげ, Chinese 中文类名, accented Ñoño) to be accepted in the New Java File and Rename wizards.

Fixes microsoft/vscode-java-dependency#789

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
 
     - name: Setup Build Environment
       run: |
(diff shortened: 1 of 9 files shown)


wenytang-ms:
before
08_type-japanese-name_after

fixed:
08_type-japanese-name_after

chagong (approved):

PR #1022, opened 2026-05-28
fix(upgrade): check extension version before calling gotoAgentMode

frankliu20:
### Why

From telemetry analysis, ~67% of `_java.upgradeWithCopilot` failures are `"command 'appmod.javaUpgrade.gotoAgentMode' not found"`. This happens because the appmod extension is installed but too old (< 1.15.0) to register the `gotoAgentMode` command. The current code only checks whether the extension is *installed*, not whether it's *up-to-date*.

Additionally, the notification button says "Upgrade Now" even when the user needs to install or update the appmod extension first — which is confusing.

### What

1. **Version check before `gotoAgentMode`** — `checkOrInstallAppModExtensionForUpgrade` now verifies the installed extension version is ≥ 1.15.0. If outdated, it triggers an update via `workbench.extensions.installExtension` and prompts the user to reload VS Code (required for the updated extension to take effect).

2. **Three-state notification UX** — Notification buttons and message body now distinguish between:
   - **up-to-date** → "Upgrade Now" / "Fix Now"
   - **outdated** → "Update Extension and Upgrade" / "Update Extension and Fix"
   - **not-installed** → "Install Extension and Upgrade" / "Install Extension and Fix"

3. **Display name update** — Renamed to "GitHub Copilot modernization" to match the extension's current marketplace name.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -36,11 +36,13 @@ export namespace ExtensionName {
     export const APP_MODERNIZATION_FOR_JAVA = "vscjava.migrate-java-to-azure";
     // Java upgrade extension is merged into app modernization extension
     export const APP_MODERNIZATION_UPGRADE_FOR_JAVA = APP_MODERNIZATION_FOR_JAVA;
-    export const APP_MODERNIZATION_EXTENSION_NAME = "GitHub Copilot app modernization";
+    export const APP_MODERNIZATION_EXTENSION_NAME = "GitHub Copilot modernization";
 }
 
 export namespace Upgrade {
     export const PACKAGE_ID_FOR_JAVA_RUNTIME = "java:*";
+    /** Minimum version of the appmod extension that supports gotoAgentMode command */
+    export const MIN_APPMOD_VERSION = "1.15.0";
 }
 
 /**
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -43,6 +43,8 @@ export namespace Upgrade {
     export const PACKAGE_ID_FOR_JAVA_RUNTIME = "java:*";
     /** Minimum version of the appmod extension that supports gotoAgentMode command */
     export const MIN_APPMOD_VERSION = "1.15.0";
+    export const SOURCE_JAVA_UPGRADE = "vscode-java-dependency.java-upgrade";
+    export const SOURCE_CVE = "vscode-java-dependency.cve";
 }
 
 /**
--- a/src/upgrade/display/notificationManager.ts
+++ b/src/upgrade/display/notificationManager.ts
@@ -1,14 +1,13 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-import { commands, ExtensionContext, extensions, window } from "vscode";
-import * as semver from "semver";
+import { commands, ExtensionContext, window } from "vscode";
 import { UpgradeReason, type IUpgradeIssuesRenderer, type UpgradeIssue } from "../type";
-import { buildCVENotificationMessage, buildFixPrompt, buildNotificationMessage, type ExtensionState } from "../utility";
+import { buildCVENotificationMessage, buildFixPrompt, buildNotificationMessage, getExtensionState } from "../utility";
 import { Commands } from "../../commands";
 import { Settings } from "../../settings";
 import { instrumentOperation, sendInfo } from "vscode-extension-telemetry-wrapper";
-import { ExtensionName, Upgrade } from "../../constants";
+import { ExtensionName } from "../../constants";
 import { CveUpgradeIssue } from "../cve";
 
 const KEY_PREFIX = 'javaupgrade.notificationManager';
@@ -27,19 +26,6 @@ const SECONDS_COUNT_BEFORE_NOTIFICATION_RESHOW = 10 * SECONDS_IN_A_DAY;
 
 function getNowTs() {
     return Number(new Date()) / 1000;
-}
-
-function getExtensionState(): ExtensionState {
-    const ext = extensions.getExtension(ExtensionName.APP_MODERNIZATION_UPGRADE_FOR_JAVA);
-    if (!ext) {
-        return "not-installed";
-    }
-    const version = ext.packageJSON?.version;
-    if (version && semver.gte(version, Upgrade.MIN_APPMOD_VERSION)) {
-        return "up-to-date";
-    }
-    // Treat missing version as outdated (conservative)
-    return "outdated";
 }
 
 class NotificationManager implements IUpgradeIssuesRenderer {
@@ -77,7 +63,7 @@ class NotificationManager implements IUpgradeIssuesRenderer {
                 }
                 this.hasShown = true;
 
-                const extensionState = getExtensionState();
+                const extensionState = getExtensionState(ExtensionName.APP_MODERNIZATION_UPGRADE_FOR_JAVA);
                 const prompt = buildFixPrompt(issue);
 
                 let notificationMessage = "";
--- a/src/upgrade/display/notificationManager.ts
+++ b/src/upgrade/display/notificationManager.ts
@@ -7,7 +7,7 @@ import { buildCVENotificationMessage, buildFixPrompt, buildNotificationMessage, 
 import { Commands } from "../../commands";
 import { Settings } from "../../settings";
 import { instrumentOperation, sendInfo } from "vscode-extension-telemetry-wrapper";
-import { ExtensionName } from "../../constants";
+import { ExtensionName, Upgrade } from "../../constants";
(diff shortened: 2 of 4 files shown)


src/upgrade/display/notificationManager.ts
```diff
@@ -26,6 +29,18 @@ function getNowTs() {
...
+function getExtensionState(): ExtensionState {
+    const ext = extensions.getExtension(ExtensionName.APP_MODERNIZATION_UPGRADE_FOR_JAVA);
+    if (!ext) {
+        return "not-installed";
+    }
+    const version = ext.packageJSON?.version;
+    if (version && semver.gte(version, Upgrade.MIN_APPMOD_VERSION)) {
+        return "up-to-date";
+    }
+    return "outdated";
```
leonard520: If version is undefined, it will go to this condition.

frankliu20: Yes, we treat it as 'out-dated'

src/upgrade/utility.ts
```diff
@@ -155,10 +168,28 @@ export async function checkOrPopupToInstallAppModExtensionForModernization(
...
+            "GitHub Copilot modernization extension has been updated. Reload VS Code to start the upgrade experience.",
+            "Reload Now"
+        );
+        if (reload === "Reload Now") {
+            await commands.executeCommand("workbench.action.reloadWindow");
+        }
+        return;
+    }
+
     await checkOrPromptToEnableAppModExtension("upgrade");
```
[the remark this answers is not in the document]
frankliu20: Fixed

src/upgrade/utility.ts
```diff
@@ -154,11 +167,35 @@ export async function checkOrPopupToInstallAppModExtensionForModernization(
...
+        } else {
+            await window.showInformationMessage(
+                `${ExtensionName.APP_MODERNIZATION_EXTENSION_NAME} extension has been updated. Please reload VS Code manually to start the upgrade experience.`
+            );
+            return false;
+        }
+    }
+
     await checkOrPromptToEnableAppModExtension("upgrade");
+    return true;
```
wenytang-ms: The fresh-install branch (when `ext` was `undefined` on entry) calls `installExtension`, then `checkOrPromptToEnableAppModExtension`, then returns `true` — and the caller in `upgradeManager.ts` immediately runs `Commands.GOTO_AGENT_MODE`. In practice, `installExtension` may resolve before the new extension host has fully activated and registered `appmod.javaUpgrade.gotoAgentMode`, which is exactly the failure mode this PR is trying to fix (just shifted from "too old" to "freshly installed and not yet active").

Suggestion: after `installExtension`, re-resolve `extensions.getExtension(...)` and verify the version meets `MIN_APPMOD_VERSION`, or check `await commands.getCommands(true)` for `Commands.GOTO_AGENT_MODE`. If the command still isn't there, return `false` so the caller doesn't blindly invoke it.

src/upgrade/utility.ts
```diff
@@ -154,11 +167,35 @@ export async function checkOrPopupToInstallAppModExtensionForModernization(
...
+
+    if (ext) {
+        // Extension was updated (not freshly installed) — reload required
+        const reload = await window.showInformationMessage(
+            `${ExtensionName.APP_MODERNIZATION_EXTENSION_NAME} extension has been updated. Reload VS Code to start the upgrade experience.`,
+            "Reload Now"
+        );
+        if (reload === "Reload Now") {
+            await commands.executeCommand("workbench.action.reloadWindow");
+            return true;
```
wenytang-ms: Returning `true` here is semantically wrong. `workbench.action.reloadWindow` tears down the extension host asynchronously — the `await` may or may not resolve before the host dies. If it does resolve, the caller will run `Commands.GOTO_AGENT_MODE` against the old extension that's about to be killed, racing the reload. Safer to `return false`: the upgrade will naturally pick up after reload when the user re-triggers it.

src/upgrade/utility.ts
```diff
@@ -154,11 +167,35 @@ export async function checkOrPopupToInstallAppModExtensionForModernization(
...
+        } else {
+            await window.showInformationMessage(
+                `${ExtensionName.APP_MODERNIZATION_EXTENSION_NAME} extension has been updated. Please reload VS Code manually to start the upgrade experience.`
+            );
+            return false;
```
wenytang-ms, line 219: The second toast is essentially redundant with the first one — if the user dismissed the "Reload Now" prompt, they already saw the same message and chose to ignore it. Showing another non-modal toast right after feels noisy.

Recommend just `return false` here without the extra `showInformationMessage`. If a persistent reminder is desired, a status-bar item would be less intrusive than a second toast.

src/upgrade/display/notificationManager.ts
```diff
@@ -26,6 +29,19 @@ function getNowTs() {
     return Number(new Date()) / 1000;
 }
 
+function getExtensionState(): ExtensionState {
```
wenytang-ms: The version-check logic here (`extensions.getExtension` + `semver.gte` against `MIN_APPMOD_VERSION`) duplicates the check in `checkOrInstallAppModExtensionForUpgrade`. Worth extracting a single helper in `utility.ts` (e.g. `getAppModExtensionState()`) and reusing it from both places so they can't drift apart when `MIN_APPMOD_VERSION` changes or the state machine evolves.

src/constants.ts
```diff
@@ -36,11 +36,13 @@ export namespace ExtensionName {
     export const APP_MODERNIZATION_FOR_JAVA = "vscjava.migrate-java-to-azure";
     // Java upgrade extension is merged into app modernization extension
     export const APP_MODERNIZATION_UPGRADE_FOR_JAVA = APP_MODERNIZATION_FOR_JAVA;
-    export const APP_MODERNIZATION_EXTENSION_NAME = "GitHub Copilot app modernization";
+    export const APP_MODERNIZATION_EXTENSION_NAME = "GitHub Copilot modernization";
```
wenytang-ms, line 39: Could you confirm with the appmod team that "GitHub Copilot modernization" is the final marketplace display name? This string is shown in user-facing notifications, so we'd want to avoid another rename round-trip soon after.
please show this change to title or PR description to make it conspicuous

wenytang-ms (commented):
Thanks for tackling this — the version check + three-state UX is a solid direction. Left a few comments inline; the fresh-install path and the return value after `reloadWindow()` look like the two issues most worth addressing before merge.

src/constants.ts
frankliu20: Yes, it's `GitHub Copilot modernization`, see https://marketplace.visualstudio.com/items?itemName=vscjava.migrate-java-to-azure

src/upgrade/display/notificationManager.ts
frankliu20: Sure, updated

src/upgrade/utility.ts
frankliu20: Sure, updated

src/upgrade/utility.ts
frankliu20: Sure, updated

src/upgrade/utility.ts
frankliu20: Thanks, updated

wenytang-ms:
One more follow-up — **telemetry coverage**.

The PR description cites a strong baseline: *~67% of `_java.upgradeWithCopilot` failures are `command 'appmod.javaUpgrade.gotoAgentMode' not found`*. The fix is targeted at exactly that failure mode, but after this lands we'll only be able to observe whether `command not found` errors drop — we won't be able to tell **why** they dropped (or didn't), or whether new failure modes emerge in their place.

Current telemetry only emits `operationName: "java.dependency.upgradeNotification.show"`. To validate this fix in production, it would help to extend `sendInfo` (here and in `checkOrInstallAppModExtensionForUpgrade`) with at least:

- `extensionState`: `up-to-date` | `outdated` | `not-installed` — distribution of users at the entry point
- `upgradeFlowResult`: e.g. `proceeded` | `install-pending-activation-timeout` | `reload-prompted-accepted` | `reload-prompted-dismissed` | `install-failed` — terminal state of `checkOrInstallAppModExtensionForUpgrade`

With those two dimensions you can answer post-release:
1. Did the 67% "command not found" slice actually disappear?
2. How many users hit the 2-second activation timeout (which currently returns `false` silently)?
3. How many users decline the reload prompt and need a second attempt?
4. What's the conversion rate from notification → successful `GOTO_AGENT_MODE`?

Without this, we're flying blind on whether the fix is fully effective and whether the magic 2s wait is the right value.

wenytang-ms (approved):

wenytang-ms (dismissed):

wenytang-ms (approved):

